PR #20, opened 2025-10-28
#13

### LINKED ISSUE #13: Trainer interfaces
**Acceptance Criteria**:
- The following interfaces exist:
  - MirrorDataset: inherits IterableDataset. `__iter__` is typed to return a token tensor and an attention mask tensor, both of shape (T).
  - MirrorModel: inherits nn.Module.
    - `training_step` takes in a batch of tokens (B, T) and a batch of attention masks (B, T), and returns a loss
    - `configure_optimizers` is typed to return an optimizer
- The above interfaces have a single, placeholder implementation each. These should be *very* basic, just returning some random value of the right shapes. Full implementations will come in later tickets.
- Trainer class is created:
  - `fit` takes in a `MirrorDataset` and a `MirrorModel`, and returns nothing. For now, just `pass`, implementation will come later.
  - is also has a `fabric` property that is a `Fabric` instance
- A trainer is created in the main method. If the `fit` subcommand is used, the `fit` method is called with a placeholder model and dataset.

Plyb:
closes #13 

How I tested: not much functionality here, but I did run the script to make sure all the imports and such were working.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,6 +6,7 @@ filelock==3.20.0
 frozenlist==1.8.0
 fsspec==2025.9.0
 idna==3.11
+jaxtyping==0.3.3
 Jinja2==3.1.6
 jsonargparse==4.42.0
 lightning==2.5.5
@@ -41,4 +42,5 @@ torchmetrics==1.8.2
 tqdm==4.67.1
 triton==3.5.0
 typing_extensions==4.15.0
+wadler_lindig==0.1.7
 yarl==1.22.0


naataaniitsosie (commented):
What's our approach for testing? Do we have any unit tests? I know you said we can manually test, but do we have any unit tests to ensure we don't break our classes and things like that?

Plyb:
> What's our approach for testing? Do we have any unit tests? I know you said we can manually test, but do we have any unit tests to ensure we don't break our classes and things like that?

Haven't gotten that far yet. In my personal opinion, things like unit tests usually end up being just a form of code duplication if they aren't handled carefully. Integration tests are better, but more expensive to run (especially in a compute constrained domain like ML). I can bring up testing in the next MIRROR team meeting.

PR #28, opened 2025-10-31
adjust architecture

### LINKED ISSUE #27: Architecture Fixes
I realized there were a couple of problems with the interfaces I originally designed.
1. It did not take into account the fact that that models expect a particular tokenization scheme.
2. IterableDataset on its own is incompatible with being able to skip a portion of the dataset, and is also incompatible with `Sampler`s. We'll likely need that functionality down the line.

**Acceptance Criteria**
- A new interface, `MirrorTokenizer` exists
  - it has abstract method(s) for taking in example strings (or a batch of strings) and returns `TokenTensor`s (or `TokenBatch`es).
  - It also has an id that uniquely identifies the tokenizer. This will be used to identify the save location for pre-tokenized data.
- Instead of returning `TokenTensor`s and `AttentionMask`s, `MirrorDataset` returns a sequence of strings (training examples). It also has a unique id that will be used for identifying the save location of pre-processed data from the dataset.
- `MirrorModel`s now have a `MirrorTokenizer`
- `MirrorTrainer`'s `fit` method is updated to use the new architecture. Complex dataset logic is stubbed and follow up tickets are created. This should just be the interface for now.

Plyb:
closes #27 

What I did to test: reran the fit script and observed no errors

### DIFF
--- a/src/tokenizers/mirror_tokenizer.py
+++ b/src/tokenizers/mirror_tokenizer.py
@@ -17,8 +17,15 @@ class MirrorTokenizer(ABC):
         pass
 
     @abstractmethod
-    def tokenize(self, text: str) -> TokenTensor:
+    def encode(self, text: str) -> TokenTensor:
         pass
 
-    def tokenize_batch(self, texts: Sequence[str]) -> TokenBatch:
-        return torch.stack([self.tokenize(text) for text in texts])
+    def encode_batch(self, texts: Sequence[str]) -> TokenBatch:
+        return torch.stack([self.encode(text) for text in texts])
+
+    @abstractmethod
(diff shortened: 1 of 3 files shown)


src/tokenizers/mirror_tokenizer.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+class MirrorTokenizer(ABC):
+    @property
+    @abstractmethod
+    def tokenization_id(self) -> str:
+        pass
+
+    @abstractmethod
+    def tokenize(self, text: str) -> TokenTensor:
```
JayOrten: I would split this up to have an "encode" and a "decode" method. Probably have a batch version of each?

PR #29, opened 2025-11-03
5 dataset selection

### LINKED ISSUE #5: Dataset Selection
**Acceptance Criteria**:
- At least two real datasets are created (implementing `MirrorDataset`), at least one of which requires a constructor parameter
- A `--data` required, named flag is added to the main script. This takes in the name of a data class. An instance of that dataset is created and passed to the fit method of the trainer.
  - If the dataset requires a constructor parameter, this would also be required as a CLI flag, like `--data.the_param`.
- The datasets read tokenized data from a subfolder of `~/nobackup/autodelete/mirror_data/tokenized`
- `PlaceholderDataset` is removed

Plyb:
closes #5 

How I tested this: Ran some training runs, observed no errors. I also had temporary print statements that verified the datasets were actually returning real example strings. For the docstrings, I also verified that `--data.help` prints the documentation.


launch.sh
```diff
@@ -9,4 +9,4 @@
 
 source env/bin/activate
 
-python src/main.py fit
+python src/main.py fit --data.class_path mirror.datasets.wikitext_dataset.WikitextDataset --data.head 10
```
Plyb: comment

src/mirror/datasets/imdb_dataset.py
```diff
@@ -0,0 +1,36 @@
...
+from mirror.datasets.mirror_dataset import MirrorDataset
+from mirror.datasets.util import load_hf_from_cache_or_download
+
+hf_dataset_path = 'stanfordnlp/imdb'
+
+
+class ImdbDataset(MirrorDataset):
+    def __init__(
+        self,
+        head: int,
```
JayOrten: What is head? Could use some comment here

JayOrten (approved):
I left some comments but it's nothing that needs to be implemented right now in this merge, mostly future cosiderations

JayOrten (approved):

PR #38, opened 2025-11-11
4 strategies distributed training

### LINKED ISSUE #4: Strategies (Distributed Training)
**Acceptance Criteria**:
- `Trainer` takes in a fabric `Strategy`, such as `FSDPStrategy`
- The strategy is used for distributed training (both single node and multi-node)
- Strategy is configurable from the CLI

**Testing Notes**:
Make sure checkpointing still works properly

Plyb:
closes #4 

What I did to test: I added a couple of print statements to verify the following when running the script on 2 nodes with 2 gpus each:
- each rank is launched (printing `get_rank`)
- each rank has different data given to it
- can load checkpoints (no errors)

I also ensured that you can do `--strategy=DDPStrategy` without errors.

### DIFF
--- a/launch.sh
+++ b/launch.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 #SBATCH --time=01:00:00   # walltime
-#SBATCH --ntasks-per-node=2   # number of processor cores (i.e. tasks)
+#SBATCH --ntasks-per-node=1   # number of processor cores (i.e. tasks)
 #SBATCH --nodes=1   # number of nodes
 #SBATCH --gpus-per-node=1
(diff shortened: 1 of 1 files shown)


src/mirror/models/placeholder_model.py
```diff
@@ -10,7 +10,7 @@
 class PlaceholderModel(MirrorModel):
     def __init__(self) -> None:
         super().__init__()
-        self.parameter = nn.Parameter(torch.tensor(0.0, device=device))
+        self.parameter = nn.Parameter(torch.tensor([0.0], device=device))
```
Plyb: required to satisfy FSDPStrategy

PR #46, opened 2025-12-04
Resolution to issue 17 jinja script

logicalpencils:
The jinja script has been written, introducing the template file slurm.jinja in the src/templates subfolder, and altering main.py to receive SLURM parameters in the CLI as well as what it accepted before. A new file, src/run.py, is now what is called by the sbatch subprocess from main.py to fit the model (and where other runtime functionality will be added).

All details for running anything is handled in the command line, e.g. 

python src/main.py fit --data="YourDataset" --time="2:13:00" --callbacks='[{"class_path": "mirror.callbacks.checkpoint_callback.CheckpointCallback", "init_args": {"every_n_train_steps": 3}}]'


logicalpencils (commented):
I'll be fixing the hardcoding issues regarding "head" and other parameters meant to be passed into an arbitrary Dataset class.

PR #54, opened 2025-12-08
45 fix requirementstxts torch version

### LINKED ISSUE #45: Fix requirements.txt's torch version
As it stands, the requirements.txt file doesn't download torch properly (you get an error saying it cant find the version). The requirements.txt file with made with pip freeze, but that included the `+cu126` suffix on the end. Ideally we'd have an easy way to update the requirements.txt file (or some equivalent) if we need to add more packages in the future, but also make sure it is usable by new people.

One thing that works is to change the torch line to be `torch==2.9.0` (remove the `+cu126` suffix), and add `--extra-index-url https://download.pytorch.org/whl/cu126` as the first line of the file (or add that as a flag when running the install command). However, that doesn't address updating the requirements file in the future.

**Acceptance Criteria**:
- must have:
  - the requirements.txt, or some equivalent, can be used to create a fresh python environment with all of the dependencies we need. This could be through added instructions in the readme, a better requirements.txt file, or changing how we're doing environments altogether.
- optional, but at least look into it and see what the options are:
  - there is some way to easily, automatically create the requirements file in the future if new packages are added.

Plyb:
closes #45 

What I did to test: following the instructions in the readme installs without issue, and running launch.sh doesn't have any import or other errors.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -3,3 +3,11 @@ channels:
   - conda-forge
 dependencies:
   - python=3.12.11
+  - pip
+  - pip:
+    - aiohttp==3.13.1
+    - datasets==4.3.0
+    - jaxtyping==0.3.3
+    - jsonargparse==4.42.0
+    - --extra-index-url https://download.pytorch.org/whl/cu126
+    - lightning==2.5.5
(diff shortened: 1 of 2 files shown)


jaded0 (approved):
fixes issue #45

PR #63, opened 2026-01-08
21 batch size and collation

### LINKED ISSUE #21: Batch size and Collation
**Acceptance Criteria**:
- batch size is configurable (via CLI or config files if those are set up already)
- batches are collated by padding the end of shorter token tensors. Attention masks mask out the padding.

JansenNye:
Ran training tests on wikitext dataset using batch sizes of 1 and 2. 
Closes #21

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,9 @@
 ## Instructions
 
 1. Create a local mamba environment `mamba create --yes -f environment.yml -p ./.env`
-1. Activate the environment `mamba activate ./.env`
-1. Install requirements `pip install -r requirements.txt`
-1. (Optional) deactivate the environment `mamba deactivate`
-1. Run using `launch.sh`. Modify `launch.sh` as needed. Later we will have a better way to launch.
-  1. Example using callbacks: `python src/main.py fit --callbacks='[{"class_path": "mirror.callbacks.checkpoint_callback.CheckpointCallback", "init_args": {"every_n_train_steps": 3}}]'`
+  a. If at some point you need to update your local environment run `mamba env update --file environment.yml --prune -p ./.env` 
+2. Activate the environment `mamba activate ./.env`
+3. (Optional) deactivate the environment `mamba deactivate`
+4. Run using `launch.sh`. Modify `launch.sh` as needed. Later we will have a better way to launch.
+  a. Example using callbacks: `python src/main.py fit --callbacks='[{"class_path": "mirror.callbacks.checkpoint_callback.CheckpointCallback", "init_args": {"every_n_train_steps": 3}}]'`
 
--- a/environment.yml
+++ b/environment.yml
@@ -11,3 +11,4 @@ dependencies:
     - jsonargparse==4.42.0
     - --extra-index-url https://download.pytorch.org/whl/cu126
     - lightning==2.5.5
+    - tqdm==4.67.1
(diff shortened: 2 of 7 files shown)


src/mirror/util.py
```diff
@@ -21,3 +22,20 @@ def assert_can_download(item_name_to_download: str):
 
 def is_power_of_ten(n: int):
     return n > 0 and math.log10(n).is_integer()
+
+def pad_to_longest(batch, pad_token):
```
Plyb: can we put some type hints on this? there's a types file that might have a type that would work for the batch.

src/mirror/util.py
```diff
@@ -21,3 +22,20 @@ def assert_can_download(item_name_to_download: str):
...
+    device = batch[0].device if torch.is_tensor(batch[0]) else torch.device("cpu")
+    lens = torch.tensor([b.numel() for b in batch], device=device, dtype=torch.long)
+    max_len = int(lens.max().item())
+    batch_size = len(batch)
+
+    tokens = torch.full((batch_size, max_len), int(pad_token), dtype=torch.long, device=device)
+
+    ar = torch.arange(max_len, device=batch[0].device)
+    attention_mask = (ar.unsqueeze(0) < lens.unsqueeze(1)).to(torch.long)
```
Plyb: Nifty! It might be worth running a test printing out the attention masks just to make sure this is doing what we expect it to.

JansenNye:
Ran training tests on wikitext dataset using batch sizes of 1 and 2 as well as randomized examples from the placeholder model

src/mirror/models/placeholder_model.py
```diff
@@ -11,8 +11,8 @@ class PlaceholderModel(MirrorModel):
     def __init__(self) -> None:
         super().__init__()
         self.parameter = nn.Parameter(torch.tensor([0.0], device=device))
-
-    _tokenizer = PlaceholderTokenizer()
+        self.pad_token_id = -1
```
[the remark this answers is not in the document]
Plyb: This still needs to be removed

src/mirror/util.py
Plyb: This also still needs to be done

src/mirror/util.py
```diff
@@ -21,3 +22,26 @@ def assert_can_download(item_name_to_download: str):
...
+    batch_size = len(batch)
+
+    tokens = torch.full((batch_size, max_len), int(pad_token), dtype=torch.long, device=dev)
+
+    ar = torch.arange(max_len, device=dev)
+    attention_mask = (ar.unsqueeze(0) < lens.unsqueeze(1)).to(torch.long)
+
+    print("raw batch[0]:", batch[0])
+    print("lens:", lens)
+    print("attention_mask:\n", attention_mask)
```
Plyb: These should be removed please

Plyb (changes requested):
Looks like some comments didn't get addressed. Maybe you just forgot to push?

src/mirror/util.py
```diff
@@ -21,3 +23,20 @@ def assert_can_download(item_name_to_download: str):
...
 def is_power_of_ten(n: int):
     return n > 0 and math.log10(n).is_integer()
+
+def pad_to_longest(batch: TokenBatch, pad_token: int):
+    dev = device
+    lens = torch.tensor([b.numel() for b in batch], device=dev, dtype=torch.long)
+    max_len = lens.max().item()
+    batch_size = len(batch)
+
+    tokens = torch.full((batch_size, max_len), int(pad_token), dtype=torch.long, device=dev)
```
Plyb: is there any particular reason to not just use `device` directly on line 29 and line 33? as opposed to making a new variable for it

Plyb (changes requested):
last thing

PR #66, opened 2026-01-14
Added a tqdm progress bar for displaying progress as we go. Adjusted …

### LINKED ISSUE #11: Progress Tracker
**Acceptance Criteria**:
- A callback exists that prints a progress bar (or equivalent information) with training loss
  - possibly using `tqdm`?
- This is a default callback
- Only the global rank 0 prints this info

morrisoncheney:
…the loss input in our callbacks to be a float, in order to troubleshoot an error that occured when calling loss.item() on our loss tensor when it came after backprop.

When the code was functionally running with a single CPU, I tested it again using 2 GPUs on the supercomputer at the same time. It successfully output the desired result. 
Tests using 3 GPUs failed, although I don't think that was related to the code that I added, but rather something to do with having 10 batches, and it wasn't divisible by 3 causing an error when the code couldn't evenly split the data amongst itself. 

Closes #11

### DIFF
--- a/src/mirror/trainer.py
+++ b/src/mirror/trainer.py
@@ -64,7 +64,7 @@ class Trainer:
         dataloader = DataLoader(preprocessed_dataset)
         dataloader = self.fabric.setup_dataloaders(dataloader, move_to_device=not is_login_node())
 
-        self.fabric.call('on_fit_start', fabric=self.fabric, model=model, optimizer=optimizer, dataset=dataset, training_run_id=training_run_id, n_batches=len(dataloader),)
+        self.fabric.call('on_fit_start', fabric=self.fabric, model=model, optimizer=optimizer, dataset=dataset, training_run_id=training_run_id, n_batches=len(dataloader))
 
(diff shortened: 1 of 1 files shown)


src/mirror/trainer.py
```diff
@@ -62,15 +64,16 @@ def fit(self, model: MirrorModel, dataset: MirrorDataset, checkpoint: Checkpoint
         dataloader = DataLoader(preprocessed_dataset)
         dataloader = self.fabric.setup_dataloaders(dataloader, move_to_device=not is_login_node())
 
-        self.fabric.call('on_fit_start', fabric=self.fabric, model=model, optimizer=optimizer, dataset=dataset, training_run_id=training_run_id)
+        self.fabric.call('on_fit_start', fabric=self.fabric, model=model, optimizer=optimizer, dataset=dataset, training_run_id=training_run_id, n_batches=len(dataloader),)
```
Plyb: nitpick: trailing comma isn't usually the style for a single line thing. They're mostly there so that git doesn't tell you you're modifying two lines if you add another line to a comma separated list.

PR #68, opened 2026-01-15
Added a command line configurable 'epochs' variable, that is called i…

### LINKED ISSUE #34: Multiple epochs
**Acceptance Criteria**:
- The trainer supports multiple epochs (times going through the full dataset)
- The number of epochs is command-line configurable
- The progress bar behaves correctly (taking into account that there will be more batches if there are more than one epochs)

morrisoncheney:
…n the fit function on main, and then is called into the fit method of our trainer, running a training loop for the number of epochs that we would like. I also adjusted the progress bar to account for the extra repetitions such that it outputs correct results.

To test this, I ran the placeholder on a variable number of epochs, with only a single compute node, and then adjusted to 5 compute nodes, and ran the model with different numbers of epochs and it continued to function as desired.

Closes #34

### DIFF
--- a/src/mirror/trainer.py
+++ b/src/mirror/trainer.py
@@ -71,7 +71,7 @@ class Trainer:
         dataloader = self.fabric.setup_dataloaders(dataloader, move_to_device=not is_login_node())
 
         self.fabric.call('on_fit_start', fabric=self.fabric, model=model, optimizer=optimizer, dataset=dataset,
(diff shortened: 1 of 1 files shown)


Plyb (changes requested):

Plyb:
@morrisoncheney Could I have you resolve the merge conflicts with main? (checkout main, pull, checkout branch, do `git merge main`, push)

PR #69, opened 2026-01-15
17 jinja script just run python

### LINKED ISSUE #17: Jinja script (just run python)
**Acceptance Criteria**:
- running the main script launches a slurm job (the `.sh` slurm script is no longer needed)
- SLURM parameters are added to the CLI for the main script
- The separate SLURM script is removed

**Testing notes**: `is_login_node` still works correctly

JansenNye:
Tested running training jobs using Wikitext dataset.
Closes #17

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -28,6 +28,9 @@ import mirror.datasets
 
 Subcommand = Literal['fit'] | Literal['test']
 
+# This is only ever assigned by the parser dump 
+# Could change to pass as parameter to main when parser is updated/changed
+run_config_yaml = ""
 
 def main(
     subcommand: Subcommand,
@@ -38,6 +41,8 @@ def main(
     callbacks: List[Callback] = [],
     checkpoint: CheckpointIdentifier | None = None,
     slurm: SlurmConfig = SlurmConfig(),
+    epochs: int = 1,
+    batch_size: int = 1,
 ):
     # These warnings happen internal to Fabric, so there's not much we can do about them.
     warnings.filterwarnings('ignore', category=FutureWarning, message='.*Please use DTensor instead and we are deprecating ShardedTensor.*')
@@ -52,7 +57,7 @@ def main(
     
     match subcommand:
         case 'fit':
-            fit(data, strategy, devices, num_nodes, callbacks, checkpoint)
+            fit(data, strategy, devices, num_nodes, callbacks, checkpoint, epochs, batch_size)
         case _:
             print(f'unimplemented subcommand: {subcommand}')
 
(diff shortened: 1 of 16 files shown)


src/mirror/templates/slurm.jinja
```diff
@@ -0,0 +1,17 @@
...
+#SBATCH --time={{ time }}
+#SBATCH --ntasks-per-node={{ ntasks_per_node }}
+#SBATCH --nodes={{ nodes }}
+#SBATCH --gpus-per-node={{ gpus_per_node }}
+#SBATCH --mem-per-cpu={{ mem_per_cpu }}
+#SBATCH --output={{ output }}
+#SBATCH --open-mode=append
+#SBATCH --signal={{ signal }}
+{% if requeue %}#SBATCH --requeue{% endif %}
+#SBATCH --chdir={{ chdir }}
```
Plyb: what does this do?

src/main.py
```diff
@@ -49,17 +57,57 @@ def fit(
...
+    context = {
+        "time": slurm.time,
+        "ntasks_per_node": slurm.ntasks_per_node,
+        "nodes": num_nodes,
+        "gpus_per_node": slurm.gpus_per_node,
+        "mem_per_cpu": slurm.mem_per_cpu,
+        "output": slurm.output,
+        "open_mode": slurm.open_mode,
+        "signal": slurm.signal,
+        "requeue": slurm.requeue,
```
Plyb: Is it possible to do a spread here, like `**slurm`? That way if we add more properties later we don't have to change this spot? Not familiar enough with Python to know.

Plyb (changes requested):
This is super clean, I love it!

src/mirror/templates/slurm.jinja
```diff
@@ -0,0 +1,17 @@
...
+
+set -euo pipefail
```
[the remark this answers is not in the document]
JansenNye: The if requeue line means that if the template variable requeue is true it does #SBATCH --requeue.
The chdir line sets the job's working directory to whatever we pass in
set -euo pipefail is a safety thing: -e means exit immediately if any command fails, -u means treat use of unset variables as an error, -o pipefail means that the pipeline fails if any command fails. Seems like it could probably be removed if you want

src/main.py
JansenNye: Seems like we can, we just have to convert slurm into a dict beforehand. I'll implement that

src/mirror/templates/slurm.jinja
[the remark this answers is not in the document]
Plyb: Yeah, unless we have a specific reason to keep it, I'd say let's remove it for simplicity.

Plyb (changes requested):

PR #73, opened 2026-01-18
33 allow local development

### LINKED ISSUE #33: Allow local development
Right now this is working only on the supercomputer. It would be nice if we were able to run this on a local machine (with or without a gpu) for testing.

**Acceptance Criteria**:
- the program can be run on a local machine with or without a gpu
- a `device` CLI parameter is added that can select between using cpu or cuda
- the program still has sensible defaults when running on the supercomputer (uses the `SLURMEnvironment`, default device is cpu on the login nodes, etc)

A couple of places off the top of my head to check:
- if `SLURMEnvironment` is already being used, it shouldn't be used locally but should be used on the supercomputer
- anywhere `is_login_node` is being called probably will need to be updated

Implementation suggestion:
- we may want some kind of globally importable `config` dictionary or mapping that is populated in `main.py` (or something that runs similarly early on). `device` can be set in that dictionary.

JansenNye:
Tested local development training runs using Wikitext data set, including specifying --device=cuda vs --device=cpu 
Closes #33

### DIFF
--- a/src/mirror/callbacks/progress_callback.py
+++ b/src/mirror/callbacks/progress_callback.py
@@ -29,6 +29,6 @@ class ProgressCallback(Callback):
             loss: float,
             **kwargs,
     ):
-        if fabric.is_global_zero and self.progress_bar is not None:
+        if Fabric.is_global_zero and self.progress_bar is not None:
             self.progress_bar.set_postfix(Loss=f"{loss:.3f}")
             self.progress_bar.update(1)


src/mirror/config.py
```diff
@@ -0,0 +1,41 @@
...
+
+import torch
+
+DeviceType = Literal['cpu', 'cuda']
+
+
+class RuntimeConfig(TypedDict):
+    device: DeviceType
+    on_slurm: bool
+    is_login_node: bool
```
Plyb: Could we maybe use an enum here to replace `on_slurm` and `is_login_node`? Since if we're not `on_slurm`, `is_login_node` will always be false, there's really only three possibilities (local, slurm-login, slurm-compute)

src/mirror/config.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+_CONFIG: RuntimeConfig | None = None
+
+
+def init_config(device: DeviceType | None = None) -> RuntimeConfig:
```
Plyb: ooo, I like this. Keep it from being globally mutable!

Plyb (changes requested):
It's looking good! Could I have you resolve the merge conflicts? Checkout `main`, do a `git pull`, then checkout your branch and do `git merge main`. You'll get a list of files that have conflicts (which you can also see in the source control sidebar on the left in vscode). Resolve the conflicts, do `git merge --continue` (or click the button to continue in the sidebar`, and repeat until there aren't any conflicts. Then rerun your tests to make sure nothing broke.

Plyb (changes requested):
Looks like the merge might have broken ProgressCallback, I'm getting an error when trying to train. Could you look into that?

PR #74, opened 2026-01-19
Implement 57 - support for different input types to model

### LINKED ISSUE #57: Support for multiple different input types for a model
**Acceptance Criteria**:
- Datasets can output different kinds of data, besides just text training examples.
  - For instance: say I had a model which was intended to be trained on text samples with emotion labels
- If at all possible, type safety is preserved. This will probably involve using generics.

JansenNye:
Tested using Wikitext dataset using batch sizes of 1 and 2. 
Closes #57


src/mirror/trainer.py
```diff
@@ -66,7 +66,7 @@ def fit(self, model: MirrorModel, dataset: MirrorDataset, checkpoint: Checkpoint
         def collate(batch):
             return pad_to_longest(batch, pad_token=model.tokenizer.pad_token_id)
 
-        preprocessed_dataset = PreprocessedDataset(dataset, model.tokenizer)
+        preprocessed_dataset = PreprocessedDataset(dataset, model.tokenizer.encode)
```
Plyb: as it stands, this still will only work for string datasets and models that take in `TokenTensor`s, because of model.tokenizer.encode's type. I'm actually a little surprised Pylance isn't giving an error here since we technically don't know if `dataset` is compatible with `model.tokenizer.encode` (EDIT: okay it looks like it does when I put it on strict mode).

Probably the easiest way to see the issue is to make `fit` generic and give the `MirrorDataset` parameter a `TypeVar` as its type argument

launch.sh
```diff
@@ -12,4 +12,4 @@
 
 mamba activate ./.env
 
-srun python src/main.py fit --data.class_path WikitextDataset --data.head 10
+srun python src/main.py fit --data.class_path WikitextDataset --data.head 10 --batch_size 2
```
Plyb: revert please

Plyb (changes requested):

src/mirror/callbacks/callback.py
```diff
@@ -41,8 +47,7 @@ def on_train_batch_end(
             model: MirrorModel,
             optimizer: Optimizer,
             loss: float,
-            tokens: TokenBatch,
-            attention_mask: AttentionMaskBatch,
+            batch: BatchT,
```
Plyb: huh... I guess none of the callbacks actually used tokens or attention_mask? If that's the case, I probably should make a ticket to just remove this.

JansenNye: Yeah, doesn't look like it

PR #75, opened 2026-01-19
Implement 60 - support for different model output types

### LINKED ISSUE #60: Support for different output types for a model
Right now all MirrorModels are configured to return just a loss. That's fine for training, but for evaluation purposes, we'll need the actual output of the model.

**Acceptance Criteria**:
- MirrorModels can output different kinds of output (text logits, emotion classification, etc)
- The loss is still in a consistent place so it can be found by the trainer
- If at all possible, type safety is preserved (probably using generics)

JansenNye:
Tested using Wikitext dataset.
Closes #60

### DIFF
--- a/src/mirror/callbacks/test_callback.py
+++ b/src/mirror/callbacks/test_callback.py
@@ -1,6 +1,6 @@
 from mirror.callbacks.callback import Callback
 
-class TestCallback(Callback): 
+class TestCallback[ProcessedT, ModelOutputT](Callback[ProcessedT, ModelOutputT]): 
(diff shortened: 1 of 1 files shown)


src/mirror/trainer.py
Plyb: Let's make the `MirrorModel`'s have a type argument (a `TypeVar`) here and other places where `MirrorModel` is used (main, callbacks, etc). That will give us better typechecking and allow us to make sure all of the different places can actually handle the shift in return type.

Plyb (changes requested):
`TestCallback` should also have the type parameters added

PR #76, opened 2026-01-19
Implement 70 - fix Pylance complaints

### LINKED ISSUE #70: Fix pylance's complaints about kwargs in callbacks
Pylance (python typechecker, installable via vscode extension) complains about how we have the parameters of our callbacks set up at the moment. `kwargs` isn't enough to satisfy it because the arguments that are before `kwargs` don't match the order in the parent `Callback` class. It looks like if we put an `*` after the `self` parameter in each case, it works, though.

Image

**Acceptance Criteria**:
- Pylance no longer produces the error described above on any callback methods

JansenNye:
Training unaffected, but tested on Wikitext dataset
Closes #70

### DIFF
--- a/src/mirror/callbacks/callback.py
+++ b/src/mirror/callbacks/callback.py
@@ -33,7 +33,14 @@ class Callback:
     ):
         pass
 
-    def on_fit_end(self, fabric: Fabric, model: MirrorModel, optimizer: Optimizer, training_run_id: str):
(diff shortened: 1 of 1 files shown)


src/mirror/callbacks/callback.py
Plyb: Looks like the * got missed here.

PR #85, opened 2026-01-30
8 preprocessing tokenized dataset cache

### LINKED ISSUE #8: Preprocessing (Tokenized dataset cache)
**Acceptance Criteria**:
- A new subcommand `preprocess` is added
  - when called, it accepts the same `--data` flag(s) that `fit` does, and downloads/tokenizes the dataset (including any other required preprocessing), putting the tokenized data in the folder where `fit` looks for them.

morrisoncheney:
Run with `subcommand: preprocess` and this will tokenize and download a tokenized dataset, such that when later run with `subcommand: fit` the pipeline will use the pre tokenized and cached dataset.
Additionally, if you set `reset_cache: True` while running with `subcommand: preprocess` the tokenized cache will reset, meaning the old tokenized data will be deleted, and a new version of the tokenized data will be downloaded.

If data has not been preprocessed, the pipeline will simply process that dataset as it trains. Including, if multiple epochs are run, it will re tokenize every piece of the dataset at each time that it comes upon it.

To test this, I ran the functions described above.

Closes #8

### DIFF
--- a/src/mirror/subcommands.py
+++ b/src/mirror/subcommands.py
@@ -47,7 +47,7 @@ def fit(
     )
 
 def preprocess(data: MirrorDataset, preprocessor: MirrorPreprocessor, slurm: SlurmConfig = SlurmConfig()) -> None:
-    if slurm.submit and is_login_node():
+    if slurm.job_type == "compute" and is_login_node():
         job_id = _submit_slurm_job(
             python_args=sys.argv[1:],
             slurm=slurm,


src/mirror/datasets/imdb_dataset.py
```diff
@@ -21,9 +22,11 @@ def __init__(
...
         super().__init__()
         ds = load_hf_from_cache_or_download(hf_dataset_path)
-        self.examples: Sequence[str] = ds[split]['text']  # pyright: ignore
-        if head:
-            self.examples = self.examples[:head]
+
+        self.ds = ds       
+        self.head = head
+        self.split = split
+        self.preprocess_ = preprocess
```
Plyb: What are these changes for?

src/mirror/datasets/mirror_dataset.py
```diff
@@ -12,3 +14,40 @@ def dataset_id(self) -> str:
     @abstractmethod
     def __getitem__(self, index: int) -> RawT:
         pass
+
+    def preprocess(self, tokenizer: MirrorTokenizer, reset_cache: bool = False): # add types 
+        dataset_id = f"{self.dataset_id}_TKID-{tokenizer.tokenization_id}".replace("/","-")
+
+        self.ds, is_cached = load_tokenized_from_cache(
```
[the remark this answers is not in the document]
Plyb: yeah, it looks like we're not actually reading `self.ds` anywhere except for the next line (which could be using a local variable) and in `is_preprocessed` (which might be a bug?)

src/mirror/datasets/util.py
```diff
@@ -49,3 +50,43 @@ def load_hf_from_cache_or_download(
...
+    else:
+        print("Dataset is not yet cached.")
+        if preprocess and dataset:
+            dataset = dataset.map(tokenizer_function)
+            print("Dataset tokenized.")
+            dataset.save_to_disk(dataset_path)
+            print("Dataset saved to disk.")
+            is_cached = True
+
+    return dataset, is_cached
```
Plyb: A couple thoughts:
- `is_cached` doesn't seem like it's doing anything super useful here, since you can infer its value from whether or not `dataset` is `None`
- It seems a bit unintuitive that this function would only return (None, False) when both of these conditions are true:
  - either the dataset wasn't cached to begin with or we cleared it
  - preprocess was set to false

I think, if I'm following your reasoning here, it's like this because this function is serving double duty. It both loads/caches the model and also checks if it has previously been cached. I kind of thing those should be two separate functions. That would get rid of the need for `preprocess`, and there's not really much shared logic between the two purposes. The path could be abstracted into a third function they both use.

src/mirror/tokenizers/mirror_tokenizer.py
```diff
@@ -10,22 +10,48 @@ class MirrorTokenizer(ABC):
...
+        """The integer ID used for padding."""
         pass
 
-    def encode_batch(self, texts: Sequence[str]) -> TokenBatch:
-        return torch.stack([self.encode(text) for text in texts])
+    @abstractmethod
+    def encode(self, text: str) -> List[int]:
+        """
+        By Leaving out torch.tensor we are making the code more saveable, 
+        and supposedly HF will wrap this with a C++ / Rust implementation to make it fast.
```
Plyb: really? It's not just that they're built in stuff has C++/Rust implementations, they'll actually compile ours?

src/mirror/tokenizers/mirror_tokenizer.py
```diff
@@ -10,22 +10,48 @@ class MirrorTokenizer(ABC):
...
+        Call this function on the name of the column containing the text of your dataset. 
+        If left blank, this defaults to 'text' which is the typical column name.
+        Additionally, you may specify the name of your output column, where the tokenized 
+        text will be stored. 
+        
+        Returns a function to be used in a ds.map(...) function call.
+        """
+        def hf_map_callable(input):
+            input['input_ids'] = self.encode(input[data_column])
+            return input
```
Plyb: it might be equivalent, I'm not sure, but an alternative here that doesn't mutate would be to just return
```
return {
  'input_ids': self.encode(input[data_column])
}
```
HF's map functions expect a dict to be returned, and it will apply all the columns of the returned dict to the old dict

src/mirror/datasets/mirror_dataset.py
```diff
@@ -12,3 +14,40 @@ def dataset_id(self) -> str:
...
+            dataset = self.ds,
+            dataset_id = dataset_id,
+            tokenizer_function = tokenizer.create_hf_map_function_(), #can do input_column
```
Plyb: this is another thing that's assuming stuff about the shape of the dataset.

src/main.py
```diff
@@ -67,11 +68,20 @@ def main(
         return
     
     match subcommand:
+        case 'preprocess':
+            preprocess(data, reset_cache)
```
Plyb: The way this is working now, you have to give all of the arguments to `preprocess` that you would to `main`. I'm wondering if it might be a better idea to just have `main` take in a subcommand, then create separate argument parsers in each of the case statements here.

src/mirror/preprocessors/mirror_preprocessor.py
```diff
@@ -1,22 +1,20 @@
...
+    def encode(self, text: str) -> List[int]:
+        """
+        By Leaving out torch.tensor we are making the code more saveable, 
+        and supposedly HF will wrap this with a C++ / Rust implementation to make it fast.
+        The above is hopefully true for encode_batch, and hf_map.
+        """
         pass
 
-    def encode_batch(self, texts: Sequence[str]) -> TokenBatch:
+    def encode_batch(self, texts: Sequence[str]) -> Sequence[ProcessedT]:
```
Plyb: for jansen: similarly, this is being dealt with in another ticket

src/main.py
```diff
@@ -69,20 +71,21 @@ def main(subcommand: Subcommand):
...
             init = parser.instantiate_classes(cfg)
 
-            dataset = init.dataset
-            model = instantiate_model(init.model, fabric=None)
+            data = init.data
+            preprocessor = init.preprocessor
+            slurm = init.slurm
 
-            preprocess(dataset, model)
+            preprocess(data, preprocessor, slurm)
```
Plyb: especially now that `model` is gone (which requires special handling), we can probably simplify this dramatically by just using `auto_parser`, or at least `add_function_arguments`

**Follow up**

src/mirror/datasets/mirror_dataset.py
```diff
@@ -1,14 +1,24 @@
...
-    @abstractmethod
-    def dataset_id(self) -> str:
-        pass
-
+    
     @abstractmethod
     def __getitem__(self, index: int) -> RawT:
         pass
+    
+    def preprocess(self, preprocessor_function) -> Sequence[TokenTensor]:
```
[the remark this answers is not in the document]
JansenNye: Looks like this didn't get addressed ^

src/mirror/datasets/mirror_dataset.py
```diff
@@ -1,14 +1,24 @@
...
+        
+        def mappable_preprocessor_function(row: dict):
+            row['input_ids'] = preprocessor_function(row)
```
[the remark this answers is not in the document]
JansenNye: Or this ^

src/main.py
JansenNye: Not addressed ^

JansenNye (changes requested):
Seems like some of Koby's comments haven't been addressed, added a few minor thoughts of my own about code structure/naming

src/mirror/datasets/mirror_dataset.py
```diff
@@ -1,14 +1,23 @@
...
     def __getitem__(self, index: int) -> RawT:
         pass
+    
+    def preprocess[ProcessedT](self, preprocessor_function: Callable[[RawT], ProcessedT]) -> Sequence[ProcessedT]:
+        
+        def mappable_preprocessor_function(row: RawT) -> dict[str, ProcessedT]:
+            return {"input_ids": preprocessor_function(row)}
+
+        self.ds = self.ds.map(mappable_preprocessor_function)
+        print("Tokenization complete.", file=stderr)
```
Plyb: Nit
```suggestion
        print("Preprocessing complete.", file=stderr)
```

src/mirror/trainer.py
```diff
@@ -95,8 +96,12 @@ def fit(
...
-
-        preprocessed_dataset = PreprocessedDataset[RawT, ProcessedT](dataset, model.preprocessor.preprocess_example)
+        
+        if do_preprocess:
+            preprocessed_dataset = dataset.preprocess(model.preprocessor.preprocess_example)
+        else:
+            preprocessed_dataset = OnDemandPreprocessedDataset[RawT, ProcessedT](dataset, model.preprocessor.preprocess_example)
+        
         dataloader = DataLoader(
             preprocessed_dataset,
```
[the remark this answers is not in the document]
morrisoncheney: Sorry about that. I have just shut out all warnings since my pylance has started doodling with red squiggles everywhere lol.

PR #87, opened 2026-02-05
7 model selection

### LINKED ISSUE #7: Model Selection
**Acceptance Criteria**:
- At least two real models are created (implementing `MirrorModel`), at least one of which requires a constructor parameter
- A `--model` required, named flag is added to the main script. This takes in the name of a model class. An instance of that model is created and passed to the fit method of the trainer.
  - If the model requires a constructor parameter, this would also be required as a CLI flag, like `--model.the_param`.
- The datasets read tokenized data from a subfolder of `~/nobackup/autodelete/mirror_data/tokenized`
- The model is initialized under the `trainer.fabric.init_module()` context
- `PlaceholderModel` is removed

JansenNye:
Ran test training runs using Wikitext dataset with LLaMa-3.2-1B (instruct and non-instruct) and GPT-2 implemented, including testing download/cache functionality and random weight initialization.
Closes #7

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -1,4 +1,4 @@
-from jsonargparse import ArgumentParser, auto_parser
+from jsonargparse import ArgumentParser
 from typing import Literal
 from inspect import signature
 import warnings
@@ -44,11 +44,10 @@ def main(subcommand: Subcommand):
 
     match subcommand:
         case 'fit':
-            parser = auto_parser(fit, as_positional=False)
-            for action in parser._actions:
-                if action.dest == "model" and hasattr(action, "sub_add_kwargs"):
-                    action.sub_add_kwargs["instantiate"] = False
-                    break
+            parser = ArgumentParser()
+            parser.add_function_arguments(fit, as_positional=False, skip={"model", "trainer"})
+            parser.add_subclass_arguments(MirrorModel, "model", required=True, instantiate=False)
+            parser.add_subclass_arguments(Trainer, "trainer", required=False, instantiate=True)
             parser.add_argument("--device", type=str, choices=["cpu", "cuda"], default=None)
             cfg = parser.parse_args(sys.argv[2:])
 
@@ -60,10 +59,20 @@ def main(subcommand: Subcommand):
 
             init_config(cfg.device)
             init_cfg = cfg.clone()
-            del init_cfg.device
             init = parser.instantiate_classes(init_cfg)
+            trainer = init.trainer or Trainer()
+            model = init.model
+
+            trainer.launch()
+            if not (is_login_node() and init.slurm.submit):
+                model = instantiate_model(model, fabric=trainer.fabric)
+
+            if is_login_node() and not init.slurm.submit:
+                print("Model downloaded/cached. Re-run on a compute node.")
+                return
+
             fit_kwargs = {name: getattr(init, name) for name in signature(fit).parameters}
-            fit(**fit_kwargs)
+            fit(**{**fit_kwargs, "model": model, "trainer": trainer})
 
         case _:
             print(f'unimplemented subcommand: {subcommand}')
@@ -72,15 +81,12 @@ def main(subcommand: Subcommand):
 def fit(
     data: MirrorDataset,
     model: MirrorModel,
-    trainer: Trainer | None = None,
+    trainer: Trainer,
     checkpoint: CheckpointIdentifier | None = None,
     slurm: SlurmConfig = SlurmConfig(),
     epochs: int = 1,
     batch_size: int = 1,
 ):
-    if trainer is None:
-        trainer = Trainer()
-
     if slurm.submit and is_login_node():
         job_id = _submit_slurm_job(
             python_args=sys.argv[1:],
@@ -90,15 +96,6 @@ def fit(
         )
         print(f"Submitted batch job {job_id}")
         return
-
-    if is_login_node():
-        model = instantiate_model(model, fabric=None)
-        print("Model downloaded/cached. Re-run on a compute node.")
-        return
-
-    trainer.launch()
-
-    model = instantiate_model(model, fabric=trainer.fabric)
 
     trainer.fit(
         model,
--- a/src/mirror/models/mirror_llama_model.py
+++ b/src/mirror/models/mirror_llama_model.py
@@ -4,7 +4,7 @@ import torch.nn as nn
 from typing import Literal
 
 from mirror.models.mirror_model import MirrorModel
-from mirror.models.model_util import build_causal_lm, ignore_id
+from mirror.models.model_util import build_causal_lm, IGNORE_ID
 from mirror.tokenizers.mirror_llama_tokenizer import MirrorLlamaTokenizer
 from mirror.types import AttentionMaskBatch, Loss, TokenBatch, TokenTensor
 from mirror.util import get_device, pad_to_longest
@@ -19,7 +19,6 @@ class MirrorLlamaModel(MirrorModel[str, TokenTensor, tuple[TokenBatch, Attention
         super().__init__()
         hf_model_name = f"meta-llama/Llama-{id}"
         self.model = build_causal_lm(hf_model_name, weights)
-        self.parameter = nn.Parameter(torch.tensor([0.0], device=get_device()))
         self._tokenizer = MirrorLlamaTokenizer(hf_model_name)
 
     @property
@@ -36,7 +35,7 @@ class MirrorLlamaModel(MirrorModel[str, TokenTensor, tuple[TokenBatch, Attention
         input_ids, attention_mask = batch
         labels = input_ids
         if attention_mask is not None:
-            labels = labels.masked_fill(attention_mask == 0, ignore_id) 
+            labels = labels.masked_fill(attention_mask == 0, IGNORE_ID) 
 
         outputs = self.model(
             input_ids=input_ids,
--- a/src/mirror/models/model_util.py
+++ b/src/mirror/models/model_util.py
@@ -8,7 +8,7 @@ from typing import Literal, Type
 from mirror.download_util import assert_can_download, mirror_data_path
 from mirror.models.mirror_model import MirrorModel
 
-ignore_id = -100 
+IGNORE_ID = -100 
 
 def load_hf_model(
         hf_model_name: str,
@@ -84,11 +84,8 @@ def instantiate_model(model: object, fabric: Fabric | None) -> MirrorModel:
     model_parser = ArgumentParser()
     model_parser.add_subclass_arguments(MirrorModel, "model", required=True, instantiate=True)
 
-    if fabric is not None:
-        def fabric_instantiator(class_type, *args, **kwargs):
-            with fabric.init_module():
-                return class_type(*args, **kwargs)
+    if fabric is None:
+        return model_parser.instantiate_classes(Namespace(model=model)).model
 
-        model_parser.add_instantiator(fabric_instantiator, MirrorModel, subclasses=True, prepend=True)
-
-    return model_parser.instantiate_classes(Namespace(model=model)).model
+    with fabric.init_module():
+        return model_parser.instantiate_classes(Namespace(model=model)).model
--- a/src/mirror/models/mirror_gpt_model.py
+++ b/src/mirror/models/mirror_gpt_model.py
@@ -4,7 +4,7 @@ import torch.nn as nn
 from typing import Literal
 
 from mirror.models.mirror_model import MirrorModel
-from mirror.models.model_util import build_causal_lm, ignore_id
+from mirror.models.model_util import build_causal_lm, IGNORE_ID
 from mirror.tokenizers.mirror_gpt_tokenizer import MirrorGPTTokenizer
 from mirror.types import AttentionMaskBatch, Loss, TokenBatch, TokenTensor
 from mirror.util import get_device, pad_to_longest
@@ -16,7 +16,6 @@ class MirrorGPTModel(MirrorModel[str, TokenTensor, tuple[TokenBatch, AttentionMa
     def __init__(self, weights: Literal["pretrained", "random"] = "pretrained") -> None:
         super().__init__()
         self.model = build_causal_lm(hf_model_name, weights)
-        self.parameter = nn.Parameter(torch.tensor([0.0], device=get_device()))
         self._tokenizer = MirrorGPTTokenizer()
 
     @property
@@ -33,7 +32,7 @@ class MirrorGPTModel(MirrorModel[str, TokenTensor, tuple[TokenBatch, AttentionMa
         input_ids, attention_mask = batch
         labels = input_ids
         if attention_mask is not None:
-            labels = labels.masked_fill(attention_mask == 0, ignore_id) 
+            labels = labels.masked_fill(attention_mask == 0, IGNORE_ID) 
 
         outputs = self.model(
             input_ids=input_ids,


src/mirror/models/model_util.py
```diff
@@ -0,0 +1,74 @@
...
+        shutil.rmtree(model_path, ignore_errors=True)
+
+    path_exists = model_path.exists()
+    weights_present = path_exists and (any((model_path / name).exists() for name in (
+                "model.safetensors", "model.safetensors.index.json",
+                "pytorch_model.bin", "pytorch_model.bin.index.json",
+            )) 
+            or any(model_path.glob("model-*.safetensors")) 
+            or any(model_path.glob("pytorch_model-*.bin"))
+    )
```
Plyb: hmm, I don't love that we have to check this manually, it seems a bit brittle. Ideally there would be some huggingface-provided function we could use. If not though, this is fine.

src/mirror/tokenizers/mirror_gpt_tokenizer.py
```diff
@@ -0,0 +1,34 @@
...
+class MirrorGPTTokenizer(MirrorTokenizer):
+    def __init__(self, hf_model_name: str) -> None:
+        self._hf_model_name = hf_model_name
+        self._tokenizer = load_hf_tokenizer_from_cache_or_download(hf_model_name)
+        if self._tokenizer.pad_token_id is None:
+            self._tokenizer.pad_token = self._tokenizer.eos_token
+
+    @property
+    def tokenization_id(self):
+        return self._hf_model_name
```
Plyb: Something to consider: the tokenization id is used to prevent duplicates when caching the tokenized dataset. If two models use the same tokenization scheme, this will cause them to have separate tokenized dataset caches, which would use more storage than needed. Is there a way we can prevent that?

src/mirror/models/mirror_llama_model.py
```diff
@@ -0,0 +1,63 @@
...
+        super().__init__()
+        hf_model_name = f"meta-llama/Llama-{id}"
+        if weights == "pretrained":
+            self.model = load_hf_model_from_cache_or_download(
+                hf_model_name,
+                model_cls=AutoModelForCausalLM,
+            )
+        elif weights == "random":
+            config = load_hf_config_from_cache_or_download(hf_model_name)
+            self.model = AutoModelForCausalLM.from_config(config)
```
Plyb: This is complicated enough that that I wonder if it merits its own helper function so that it isn't duplicated between models. Borderline for me.

src/mirror/util.py
```diff
@@ -44,3 +59,19 @@ def pad_to_longest(batch: list[TokenTensor], pad_token: int) -> tuple[TokenBatch
         tokens[i,:L] = b
     
     return tokens, attention_mask
+
+def instantiate_model(model: object, *, fabric):
```
Plyb: could give fabric a type, and the return type of the function here.

Plyb: `jsonargparse` provides an `instantiate_classes` function. Could we use that instead of making our own implementation?

src/main.py
```diff
@@ -29,16 +30,20 @@
 import lightning.fabric.strategies
 import mirror.callbacks
 import mirror.datasets
+import mirror.models.mirror_gpt_model
+import mirror.models.mirror_llama_model
```
Plyb: Instead of doing this, create an `__init__.py` file under `models` and import that, like we're doing with `datasets` and `callbacks`.

src/main.py
```diff
@@ -131,13 +142,30 @@ def _submit_slurm_job(*, python_args: list[str], slurm: SlurmConfig, num_nodes:
...
     if hasattr(cfg, 'config'):
         del cfg.config  # pyright: ignore
 
     init = parser.instantiate_classes(cfg)
-
-    main(**init)
+    if isinstance(init, dict):
+        main(**init)
+    else:
+        main(**vars(init))
```
Plyb: So, overall with this, here are my thoughts:
- the way our main script is currently set up (pre-this ticket) makes the model instantiation tricky
  - We need to defer instantiating the model until we have fabric set up so we can do it under `init_args`
  - But also we need to have model instantiated by the time we call main, or at least fabric
  - The main function is also set up in kind of a dumb way (my bad), because not all subcommands need all the parameters of all the other ones.
- My thoughts to fix it:
  - Change the `main` function so that it *only* takes in subcommand
    - then, within each case of the match statement, create a second parser for the parameters needed for *just that subcommand* (so the parameters of the `fit` function for the "fit" subcommand in this case)
  - Modify `fit`, so that instead of taking in `strategy`, `devices`, `num_nodes`, and `callbacks`, it takes in a `Trainer` instance.
  - Inside the the "fit" subcommand's case, do something a bit like what you're doing in the `__main__` script here:
    - add all of `fit`'s parameters *except* `model`
    - add the model separately so that we don't have to instantiate it when we do `instantiate_classes`.
    - instantiate everything except the model (note that this will include the trainer)
    - use the trainer to instantiate the model under `init_module`.
    - pass everything into `fit`

I'm considering whether or not we should do the first part (breaking up the parameter parsing into the case branches) in its own ticket, since that will be useful for #8, and we can get that out first. If that's too much of a pain though, we can just leave it in here.

src/mirror/models/mirror_llama_model.py
```diff
@@ -0,0 +1,63 @@
...
+
+    def preprocess_example(self, text: str) -> TokenTensor:
+        return self.tokenizer.encode(text)
+
+    def collate(self, examples: list[TokenTensor]) -> tuple[TokenBatch, AttentionMaskBatch]:
+        return pad_to_longest(examples, pad_token=self.tokenizer.pad_token_id)
+
+    def training_step(self, batch: tuple[TokenBatch, AttentionMaskBatch]) -> Loss:
+        input_ids, attention_mask = batch
+        labels = input_ids.clone()
```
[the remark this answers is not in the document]
JansenNye: We need to make labels separate from input_ids because we then modify labels to set padded positions to -100 and we don't want to set input_ids to -100 too

src/mirror/models/mirror_llama_model.py
```diff
@@ -0,0 +1,63 @@
...
+        if attention_mask is not None:
+            labels = labels.masked_fill(attention_mask == 0, -100)
```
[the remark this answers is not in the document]
JansenNye: attention_mask only tells the transformer what tokens to attend to - it doesn't stop the loss function from calculating loss at those positions, so if we didn't do this we'd calculate loss on padding tokens and basically be training the model to predict padding.
-100 is transformers' ignore_index; it's doing what the attention mask doesn't, telling the loss function not to calculate loss on these positions.

src/mirror/models/model_util.py
```diff
@@ -0,0 +1,74 @@
...
+from mirror.util import assert_can_download, mirror_data_path
+
+
+def load_hf_model_from_cache_or_download(
+        hf_model_name: str | None = None,
+        reset_cache: bool = False,
+        model_cls: Type[PreTrainedModel] = AutoModel,
+) -> PreTrainedModel:
+    """
+    The first time this is called with a particular path/name pair, it will download
```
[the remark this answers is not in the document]
JansenNye: Ah, yeah, forgot to change the docstring

src/main.py
```diff
@@ -65,16 +70,23 @@ def main(
         job_id = _submit_slurm_job(python_args=sys.argv[1:], slurm=slurm, num_nodes=num_nodes, devices=devices)
         print(f"Submitted batch job {job_id}")
         return
+
+    if is_login_node() and not slurm.submit and subcommand == "fit":
```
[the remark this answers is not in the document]
JansenNye: I initially had it implemented this way and it wasn't downloading properly, but I can't imagine why it wouldn't work so I'll try again

src/mirror/models/model_util.py
```diff
@@ -14,61 +14,63 @@ def load_hf_model_from_cache_or_download(
...
-    config.save_pretrained(model_path)
-    return config
+    # Avoid creating config-only directory 
+    load_hf_model_from_cache_or_download(hf_model_name=hf_model_name, reset_cache=reset_cache)
+
+    return AutoConfig.from_pretrained(model_path, local_files_only=True)
+
+def build_causal_lm(
+        model_name: str | None = None,
+        weights: str = "pretrained"
```
Plyb: we can be a bit more specific here by saying `weights: Literal["pretrained", "random"]` (which is safer too, otherwise putting in a different string would result in returning `None`. Or maybe an exception, not sure)

src/mirror/models/model_util.py
```diff
@@ -14,61 +14,63 @@ def load_hf_model_from_cache_or_download(
...
-    model_path.mkdir(parents=True, exist_ok=True)
-    config.save_pretrained(model_path)
-    return config
+    # Avoid creating config-only directory 
+    load_hf_model_from_cache_or_download(hf_model_name=hf_model_name, reset_cache=reset_cache)
+
+    return AutoConfig.from_pretrained(model_path, local_files_only=True)
+
+def build_causal_lm(
+        model_name: str | None = None,
```
Plyb: Why allow `None` here?

src/mirror/tokenizers/mirror_gpt_tokenizer.py
```diff
@@ -0,0 +1,34 @@
...
+from mirror.tokenizers.tokenizer_util import load_hf_tokenizer_from_cache_or_download
+from mirror.util import get_device
+
+
+class MirrorGPTTokenizer(MirrorTokenizer):
+    def __init__(self, hf_model_name: str) -> None:
+        self._hf_model_name = hf_model_name
+        self._tokenizer = load_hf_tokenizer_from_cache_or_download(hf_model_name)
+        if self._tokenizer.pad_token_id is None:
+            self._tokenizer.pad_token = self._tokenizer.eos_token
```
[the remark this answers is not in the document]
Plyb: Ah, didn't realize it had both. But `pad_token_id` is the one being checked in the `if` condition, and the one that's actually being used below. I imagine we do want to set the `pad_token_id` at least at some point.

src/mirror/tokenizers/mirror_gpt_tokenizer.py
```diff
@@ -1,34 +1,36 @@
...
-        return self._hf_model_name
+    def tokenization_id(self) -> str:
+        return "MirrorGPTTokenizer"
 
-    def encode(self, text):
+    def encode(self, text: str) -> TokenTensor:
         ids = self._tokenizer.encode(text, add_special_tokens=True)
         if len(ids) < 2:
             eos = self._tokenizer.eos_token_id
             eos = int(eos)
```
Plyb: is `eos_token_id` ever anything *but* an int?

src/mirror/tokenizers/tokenizer_util.py
```diff
@@ -1,33 +1,32 @@
...
+tokenizers_path = mirror_data_path / "tokenizers"
 
 
-def load_hf_tokenizer_from_cache_or_download(
+def load_hf_tokenizer_from_cache_or_download[HfTokenizerT: PreTrainedTokenizerBase](
         hf_model_name: str | None = None,
         reset_cache: bool = False,
-        tokenizer_cls: Type[PreTrainedTokenizerBase] = AutoTokenizer,
-) -> PreTrainedTokenizerBase:
+) -> HfTokenizerT:
```
Plyb: python is weird with its generics. You can't explicitly pass a type argument to a function, the only way is to have it be inferred from the (value) parameters. So in this case, `HfTokenizerT` is always just going to be the most general type allowed (`PreTrainedTokenizerBase`). Might as well remove the type parameter.

Plyb (changes requested):
A few comments from last time weren't addressed. One thing I usually do before sending something into review is to review the PR myself first. That tends to catch that kind of thing. I've resolved all the comments that were fixed. Overall great work! I'm liking it.

src/mirror/util.py
JansenNye: Yes, it's just a little tricky because of initializing the model inside the fabric.trainer.init_module() context, but I'll implement that

src/main.py
```diff
@@ -65,16 +70,23 @@ def main(
...
+        trainer = Trainer(strategy, devices, num_nodes, callbacks)
+        trainer.launch()
```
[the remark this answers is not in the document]
JansenNye: Gotcha, I'll separate the login node/model download flow from the actual fitting one

src/mirror/models/model_util.py
JansenNye: Yeah, that wouldn't ever be a valid input, I'll remove it

src/mirror/models/model_util.py
```diff
@@ -1,16 +1,17 @@
...
+from jsonargparse import ArgumentParser, Namespace
 from lightning import Fabric
 from transformers import AutoConfig, AutoModel, AutoModelForCausalLM, PreTrainedModel, PretrainedConfig
-from typing import Type
+from typing import Literal, Type
 
 from mirror.download_util import assert_can_download, mirror_data_path
 from mirror.models.mirror_model import MirrorModel
 
+ignore_id = -100
```
Plyb: minor style thing, I think it's more common to have constants be UPPER_SNAKE_CASE

src/main.py
```diff
@@ -56,16 +59,11 @@ def main(subcommand: Subcommand):
...
-                trainer=init.trainer,
-                checkpoint=init.checkpoint,
-                slurm=init.slurm,
-                epochs=init.epochs,
-                batch_size=init.batch_size,
-            )
+            init_cfg = cfg.clone()
+            del init_cfg.device
+            init = parser.instantiate_classes(init_cfg)
+            fit_kwargs = {name: getattr(init, name) for name in signature(fit).parameters}
```
Plyb: I think you only need either the `del` or the `signature` thing, not both, unless there's something I'm missing.

src/main.py
```diff
@@ -57,40 +42,72 @@ def main(
...
+            run_config_yaml = f"subcommand: fit\n{parser.dump(cfg)}"
+
+            if hasattr(cfg, 'config'):
+                del cfg.config  # pyright: ignore
+
+            init_config(cfg.device)
+            init_cfg = cfg.clone()
+            del init_cfg.device
+            init = parser.instantiate_classes(init_cfg)
+            fit_kwargs = {name: getattr(init, name) for name in signature(fit).parameters}
```
Plyb: I think we only need either the `del` or the `signature` thing, not both.

src/main.py
```diff
@@ -57,40 +42,72 @@ def main(
...
+    if is_login_node():
+        model = instantiate_model(model, fabric=None)
+        print("Model downloaded/cached. Re-run on a compute node.")
+        return
 
     trainer.launch()
 
-    with trainer.fabric.init_module():
-        model = PlaceholderModel()
+    model = instantiate_model(model, fabric=trainer.fabric)
```
Plyb: We should move the model instantiation stuff out of `fit`. As it stands, `fit` *says* it's accepting a `MirrorModel`, when it's actually accepting a namespace/dict that defines how to construct one.

Up in main, after we've done `instantate_classes`, `init.trainer.fabric` has a fabric instance which you can use to instantiate the model under `.init_module()`. After that is done, you can pass the fully initialized model in with all the other parameters to `fit`.

src/mirror/models/model_util.py
```diff
@@ -79,15 +80,15 @@ def build_causal_lm(
...
+    model_parser.add_subclass_arguments(MirrorModel, "model", required=True, instantiate=True)
+
+    if fabric is not None:
+        def fabric_instantiator(class_type, *args, **kwargs):
+            with fabric.init_module():
+                return class_type(*args, **kwargs)
+
+        model_parser.add_instantiator(fabric_instantiator, MirrorModel, subclasses=True, prepend=True)
+
+    return model_parser.instantiate_classes(Namespace(model=model)).model
```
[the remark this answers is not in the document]
JansenNye: I thought we had to do the add_instantiator, but it seems you're right, I'll implement that

PR #115, opened 2026-02-19
Remove redundant RequeueCallback

### LINKED ISSUE #81: Remove redundant Requeue callback from callbacks list
There are two requeue callbacks in Trainer's default callback list

**Acceptance Criteria**:
- There is only one requeue callback in the default callback list.

JansenNye:
Tested using standard Wikitext LLaMa training runs 
Closes #81


Plyb (approved):
Next time we should make sure we split tickets like this up. Way too complicated.

PR #126, opened 2026-02-24
Fix login node training runs

### LINKED ISSUE #77: Fix "fit" on login node (for testing)
It seems like somewhere along the way, the `fit` subcommand stopped working on the login node (it raises errors about not having a GPU available). We should be able to run `fit` just using the cpu as the device, for quick testing and downloading. I believe you can trigger a login-node only run by passing the slurm.submit=False flag.

**Acceptance Criteria**:
- running `fit` on the login node works (uses `"cpu"` as the device).

JansenNye:
Tested by running Llama-3.2-1B Wikitext training runs on the local node with slurm.job_type="local".
Closes #77

### DIFF
--- a/src/mirror/slurm_util.py
+++ b/src/mirror/slurm_util.py
@@ -1,7 +1,7 @@
 import os
 import re
 from dataclasses import dataclass
-from typing import Optional
+from typing import Optional, Literal
 
 def get_job_id():
     array_job_id = os.getenv('SLURM_ARRAY_JOB_ID')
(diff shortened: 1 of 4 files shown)


src/mirror/slurm_util.py
```diff
@@ -17,7 +17,7 @@ def get_job_id():
 
 @dataclass
 class SlurmConfig:
-    submit: bool = True
+    job_type: Literal["compute", "local", "local-download"] = "compute"
```
morrisoncheney: Kind of confusing that this is called `"local"` just because it is on the login nodes, and not our local devices. This could be fine, but may be worth changing.

PR #136, opened 2026-03-02
Preliminary implementation of overfit test

### LINKED ISSUE #30: Basic sanity test: can we overfit on a small dataset with a small model?
**Acceptance Criteria**:
- There is a reasonably computationally cheap test that makes sure we can overfit a small model on a small dataset without any errors
- We have checked if would be feasible for us to run such a test on the GitHub CI/CD. If it is, the test is set up to run for each PR.

JansenNye:
Testing by making this pull request.
Closes #30

### DIFF
--- a/.github/workflows/overfit-pr.yml
+++ b/.github/workflows/overfit-pr.yml
@@ -33,6 +33,7 @@ jobs:
           WANDB_MODE: offline
           HF_TOKEN: ${{ secrets.HF_TOKEN }}
           HUGGINGFACE_HUB_TOKEN: ${{ secrets.HF_TOKEN }}
+          MIRROR_DATA_PATH: ${{ github.workspace }}/.mirror_data
           MIRROR_PRINT_STEP_LOSS: "true"
         run: |
           python scripts/check_overfit.py --config overfit_test_config.yaml
--- a/src/mirror/models/model_util.py
+++ b/src/mirror/models/model_util.py
@@ -6,8 +6,8 @@ from transformers import AutoConfig, AutoModel, AutoModelForCausalLM, PreTrained
 from typing import Literal, Type
 
 from mirror.download_util import assert_can_download
-from mirror.util import mirror_data_path
 from mirror.models.mirror_model import MirrorModel
+from mirror.util import mirror_data_path
 
 IGNORE_ID = -100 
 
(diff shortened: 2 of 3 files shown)


.github/workflows/overfit-pr.yml
```diff
@@ -0,0 +1,34 @@
+name: PR Overfit Check
+
+on:
+  pull_request:
```
Plyb: Correct me if I'm wrong, but iirc this runs every time the PR is updated (including new pushes). This could get pretty pricey pretty fast. By default, let's only run this when a new PR is opened (or a draft is set as ready)

.github/workflows/overfit-pr.yml
```diff
@@ -0,0 +1,34 @@
+name: PR Overfit Check
```
Plyb: question: will this run even for PRs by people not in the dragn labs organization? If so we should probably prevent that so other people can't run down our api key

scripts/check_overfit.py
```diff
@@ -0,0 +1,59 @@
...
+    final_loss = losses[-1]
+    absolute_drop = initial_loss - final_loss
+    relative_drop = absolute_drop / initial_loss if initial_loss > 0 else 0.0
+
+    print(
+        f"Overfit check: initial_loss={initial_loss:.6f}, final_loss={final_loss:.6f}, "
+        f"absolute_drop={absolute_drop:.6f}, relative_drop={relative_drop:.3f}"
+    )
+
+    if absolute_drop < args.min_absolute_drop or relative_drop < args.min_relative_drop:
```
Plyb: Shouldn't this be flipped?

src/mirror/download_util.py
```diff
@@ -6,7 +6,9 @@
 
 from mirror.config import RuntimeEnvironment, get_config
 
-mirror_data_path = Path(f"/home/{os.environ['USER']}/nobackup/autodelete/mirror_data")
+mirror_data_path = Path(
+    os.getenv("MIRROR_DATA_PATH", f"/home/{os.environ['USER']}/nobackup/autodelete/mirror_data")
+)
```
Plyb: why is this needed?

Plyb (changes requested):
I'm curious what the outcome of the tokenizer downloading discussion was. Did we just not need to download it?

PR #137, opened 2026-03-02
Remove encode methods/pad_token_id, add MirrorProcessor type args

### LINKED ISSUE #130: Further Preprocessor Corrections
**Acceptance Criteria**:
- none of the preprocessors have an `encode` method anymore.
- Every usage of `MirrorPreprocessor` should be given type arguments
- `pad_token_id` is removed from `MirrorPreprocessor` (it can stay on the subclasses that need it, it's just not a concept that needs to be on the generic version anymore)

JansenNye:
Tested using standard LLaMa Wikitext training runs
Closes #130

### DIFF
--- a/.github/workflows/overfit-pr.yml
+++ b/.github/workflows/overfit-pr.yml
@@ -1,0 +1,39 @@
+name: PR Overfit Check
+
+on:
+  pull_request:
+    types: [opened, ready_for_review]
+
+jobs:
+  overfit-check:
+    runs-on: ubuntu-latest
+    timeout-minutes: 45
+    if: |
+      github.event.pull_request.author_association == 'OWNER' ||
+      github.event.pull_request.author_association == 'MEMBER' ||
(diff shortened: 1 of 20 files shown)


src/mirror/preprocessors/placeholder_preprocessor.py
```diff
@@ -5,10 +5,9 @@
...
 
-class PlaceholderPreprocessor(MirrorPreprocessor):
-    def encode(self, text):
-        return torch.tensor([1, 2, 3, 4], device=get_device())
-
+class PlaceholderPreprocessor(
+    MirrorPreprocessor[TextRow, TokenTensor, tuple[TokenBatch, AttentionMaskBatch]]
+):
     def preprocess_example(self, example: TextRow) -> TokenTensor:
         return self.encode(example['text'])
```
Plyb: the auto type checker did something right! This will break the placeholder preprocessor.

PR #154, opened 2026-03-09
95 txtdataset

### LINKED ISSUE #95: TxtDataset
Some of our data isn't available on the Huggingface hub. In particular, we have the english split Church Translation dataset, which was put together by the MATRIX lab here at BYU, which is just a big `.txt` file. Because we don't have authorization to publish that dataset, we can't put it directly in the repo. However, we *can* load the txt file onto the supercomputer and have the pipeline read from it.

**Acceptance Criteria**:
- A `TxtDataset` implementation of `MirrorDataset` is created
  - it takes a file path as a constructor parameter, and loads a `.txt` file as the data source, splitting on newlines
- Keep an eye on memory. Can we load the whole thing in one go? If not, can we load in chunks?

**The Church Translation Dataset**:
- We're using the Church Translation dataset from BYU's MATRIX lab. It contains all of the translated material from the Gospel Library app and churchofjesuschrist.org.
  - We'll only be using the English split
  - It is about 180MB
- Contact @Plyb for access to the data. We don't currently have authorization to share it outside of BYU, so it isn't published anywhere. Currently it's just a big .txt file.

JansenNye:
Tested using first 100 lines of Church Text Dataset using Llama-3.2-1B-Instruct
Closes #95


src/mirror/datasets/txt_dataset.py
```diff
@@ -0,0 +1,45 @@
...
+        self._file_path = Path(file_path)
+        self._lines: list[str] = []
+
+        with open(self._file_path, "r", encoding="utf-8") as f:
+            for i, line in enumerate(f):
+                if head is not None and i >= head:
+                    break
+                stripped = line.rstrip("\n")
+                if stripped:
+                    self._lines.append(stripped)
```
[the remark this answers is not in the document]
Plyb: Yeah, I think we should try to figure out how to get this into a huggingface dataset. As it stands, the `preprocess` function actually _requires_ a huggingface dataset on `self.ds` (which isn't currently made clear by `MirrorDataset`. See #155), so that's one thing. The other thing I'm concerned about with the implementation as it stands is that it's loading the entire dataset into memory. With large datasets, that's going to be infeasible. Afaik, huggingface has some fancy under the hood stuff to make sure it's only loading into memory what it needs (called "memory mapping" iirc). If we can get that for free by using a hf dataset that would be great. Otherwise, we'll probably need to come up with our own memory-mapping-like solution. That is, unless you've already tried this with the full church translation dataset @JansenNye?

src/mirror/datasets/wikitext_dataset.py
```diff
@@ -12,9 +12,9 @@
 
 class WikitextDataset(MirrorDataset[TextRow]):
     def __init__(
-        self,
-        head: int | None = None,
-        split: Literal['train'] | Literal['validation'] | Literal['test'] = 'train',
+            self,
+            head: int | None = None,
+            split: Literal['train'] | Literal['validation'] | Literal['test'] = 'train',
```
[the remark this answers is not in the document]
Plyb: editors seem to disagree on this. I have no opinion. Probably good to avoid unnecessary changes in PRs though to reduce merge conflicts.

src/mirror/datasets/txt_dataset.py
```diff
@@ -0,0 +1,45 @@
...
+
+    @property
+    def dataset_id(self) -> str:
+        return str(self._file_path)
```
Plyb: `dataset_id` was removed in an earlier ticket, you can get rid of this.

src/mirror/datasets/txt_dataset.py
```diff
@@ -0,0 +1,45 @@
...
+    def __getitem__(self, index: int) -> TextRow:
+        return TextRow(text=self._lines[index])
+
+    def __len__(self) -> int:
+        return len(self._lines)
+
+    def preprocess[ProcessedT](self, preprocessor_function: Callable[[TextRow], ProcessedT]) -> Sequence[ProcessedT]:
+        results = [preprocessor_function(self[i]) for i in range(len(self))]
+        print("Preprocessing complete.", file=stderr)
+        return results
```
Plyb: Oh, I see you already dealt with `preprocess`. I think I still prefer the hf dataset solution if possible. Overwriting base class methods is a code smell imo. If we can't do the hf solution, we maybe should look into ways to rearchitect this preprocess function, maybe making it independent of the MirrorDataset class and just passing a mirror dataset in. If we end up needing to go down that path, let's talk, since it's probably outside the scope of this ticket.

src/mirror/models/whitebox_transformers/whitebox_transformers.py
```diff
@@ -1,19 +1,19 @@
...
-RawHFOutputTypes = CausalLMOutputWithPast | CausalLMOutputWithCrossAttentions
-Logits = torch.FloatTensor
-LossPresent = torch.FloatTensor
-HiddenStatesPresent = List[torch.FloatTensor]
-AttentionsPresent = List[torch.FloatTensor]
+type RawHFOutputTypes = CausalLMOutputWithPast | CausalLMOutputWithCrossAttentions
+type Logits = torch.FloatTensor
+type LossPresent = torch.FloatTensor
+type HiddenStatesPresent = list[torch.FloatTensor]
+type AttentionsPresent = list[torch.FloatTensor]
```
Plyb: seems unrelated to the main ticket?

Plyb (changes requested):

PR #157, opened 2026-03-09
89 do we need pad token to longest or can we just use hf tokenizers

### LINKED ISSUE #89: Do we need `pad_token_to_longest` or can we just use HF tokenizers?
`pad_token_to_longest` might be duplicating what hf tokenizers were already doing.

**Acceptance Criteria**:
- either:
  - replace the usages of `pad_token_to_longest` with hf tokenizer calls
  - explain why `pad_token_to_longest` does something that hf tokenizers can't

JansenNye:
Tested using standard Wikitext Llama runs
Closes #89

### DIFF
--- a/src/mirror/preprocessors/mirror_gpt_preprocessor.py
+++ b/src/mirror/preprocessors/mirror_gpt_preprocessor.py
@@ -30,4 +30,4 @@ class MirrorGPTPreprocessor(
 
     @property
     def pad_token_id(self) -> int:
-        return cast(int, self._tokenizer.pad_token_id)
+        return int(self._tokenizer.pad_token_id)  # type: ignore[arg-type]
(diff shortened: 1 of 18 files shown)


src/mirror/preprocessors/mirror_gpt_preprocessor.py
```diff
@@ -1,10 +1,12 @@
+from typing import cast
+
 import torch
 from transformers import PreTrainedTokenizerBase
 
 from mirror.preprocessors.mirror_preprocessor import MirrorPreprocessor
 from mirror.preprocessors.preprocessor_util import load_hf_tokenizer
-from mirror.types import TokenTensor, TokenBatch, AttentionMaskBatch
-from mirror.util import get_device, pad_to_longest
+from mirror.types import TokenBatch, AttentionMaskBatch
```
Plyb: `TokenTensor` is being used as a type arg to `MirrorPreprocessor`

PR #169, opened 2026-03-10
Add qos slurm flag

### LINKED ISSUE #151: Add "qos" to SlurmConfig and template
A pretty common slurm flag we use is "qos". We should be able to include that in the mirror slurm config.

**Acceptance Criteria**:
- "qos" is added to both SlurmConfig and the jinja template so that a user can select a qos when they submit a job.

JansenNye:
Tested using standard Wikitext/Llama training runs, with and without `qos` slurm parameter.
Closes #151


morrisoncheney (commented):
Looks good to me

PR #188, opened 2026-03-13
141 validation and test set

### LINKED ISSUE #141: Validation and test set
**Acceptance Criteria**:
- training can have a validation and test set
- The frequency of validation checks is configurable

Some suggestions:
- A couple options for setting what should be the val/test sets
  - have a totally separate parameter on the `fit` subcommand for each
    - simple
    - need some way of making sure it doesn't overlap with the training set. Easy way to do this to start would be have a `skip` parameter on the dataset constructors in addition to the `head`. The two combined would effectively allow us to set the range when making a dataset. However, this doesn't deal with shuffling. Maybe do shuffling after `skip` and `head`.
  - Rearchitect `MirrorDataset` so that it has separate get_item functions for the train/val/test sets
- Lightning defaults to running the validation set once per epoch, and let's you set the frequency of epochs per val run (so like frequency=0.5 runs a val run every half an epoch). This seems sensible to me, but I don't really have a strong preference for how it works as long as we can set the frequency somehow.

Note: if test sets are adding too much complexity, we can split test and val into two different tickets.

JansenNye:
Tested using config file: 
```
data:
  class_path: WikitextDataset
  init_args:
    split: train
    head: 10

val_data:
  class_path: WikitextDataset
  init_args:
    split: validation
    head: 5

val_check_interval: 1.0

test_data:
  class_path: WikitextDataset
  init_args:
    split: test
    head: 5

# rest same as standard LLaMa training runs, except epochs = 3
```
Closes #141

### DIFF
--- a/src/mirror/trainer.py
+++ b/src/mirror/trainer.py
@@ -108,21 +108,6 @@ class Trainer[RawT, ProcessedT, BatchT, ModelOutputT]:
             }
             self.fabric.load(checkpoint.path, state)
 
-        if do_preprocess:
-            preprocessed_dataset = dataset.preprocess(preprocessor.preprocess_example)
-        else:
-            preprocessed_dataset = OnDemandPreprocessedDataset[RawT, ProcessedT](dataset, preprocessor.preprocess_example)
-
-        dataloader = DataLoader(
-            preprocessed_dataset,  # type: ignore[arg-type]
-            batch_size=batch_size,
-            collate_fn=preprocessor.collate,
-            drop_last=False,
-        )
-        dataloader = self.fabric.setup_dataloaders(dataloader, move_to_device=self.config['device'] == 'cuda')
-        self.fabric.call('on_fit_start', fabric=self.fabric, model=model, optimizer=optimizer, dataset=dataset,
-            training_run_id=training_run_id, n_batches=len(dataloader), epochs=epochs, run_config_yaml=run_config_yaml)
-
         dataloader = self._make_dataloader(dataset, model, batch_size, do_preprocess)
 
         val_dataloader = None


src/main.py
```diff
@@ -36,9 +37,11 @@ def main(subcommand: Subcommand):
         case 'fit':
             parser = ArgumentParser()
             parser.add_argument("--config", action=ActionConfigFile)
-            parser.add_function_arguments(fit, as_positional=False, skip={"model", "trainer", "run_config_yaml"})
+            parser.add_function_arguments(fit, as_positional=False, skip={"model", "trainer", "run_config_yaml", "val_data", "test_data"})
             parser.add_subclass_arguments(MirrorModel, "model", required=True, instantiate=False)
             parser.add_subclass_arguments(Trainer, "trainer", required=False, instantiate=True)
+            parser.add_subclass_arguments(MirrorDataset, "val_data", required=False, instantiate=True)
+            parser.add_subclass_arguments(MirrorDataset, "test_data", required=False, instantiate=True)
```
Plyb: Why do we need to treat `val_data` and `test_data` specially? Couldn't we just initialize them in the same way as `data`?

src/mirror/trainer.py
```diff
@@ -180,7 +181,10 @@ def _eval_loop(self, model, dataloader) -> float:
                 total_loss += loss.item()
                 n_batches += 1
         model.train()
-        return total_loss / n_batches if n_batches > 0 else 0.0
+        if n_batches == 0:
+            warnings.warn("Validation dataloader yielded 0 batches; returning loss 0.0")
```
Plyb: `Validation` here might be confusing if it happens with the test test

src/mirror/subcommands.py
```diff
@@ -44,6 +47,9 @@ def fit(
         batch_size,
         do_preprocess,
         run_config_yaml,
+        val_dataset=val_data,
+        test_dataset=test_data,
+        val_check_interval=val_check_interval,
```
Plyb: type mismatch here

PR #223, opened 2026-03-25
ZeroMLP example intervention

### LINKED ISSUE #79: Intervention System
**Acceptance Criteria**:
- An intervention system exists that meets the design constraints in #52 
- Interventions can be selected via configuration, similar to models, datasets, and callbacks
- Interventions are applied according to the architecture below
- A sample intervention exists for testing
  - example one you could do: take the output of all the MLPs in a model and force their output to be filled with zero.

Note: it's possible that really the only thing this ticket needs to do is build the example intervention

# Architecture

Currently we have `MirrorModel`s as one of the inputs of our training process, and we need to get a modified `MirrorModel` at the end.

An intervention should effectively be a "wrapper" for a `MirrorModel` that is *itself* a `MirrorModel`. The only real difference between a model and an intervention and a model, is that an intervention requires a model as a constructor parameter. As an example, you could have a (somewhat silly) intervention that multiplies the loss of a model by a parameter 'multiplier':
```python
class MultiplierIntervention(MirrorModel):
  def __init__(target: MirrorModel, multiplier: int):
    self.target = target
    self.multipler = multiplier

  def training_step(batch):
    return self.multiplier * self.target(batch)

  # other methods just forward to target, see below
```

Interventions should follow encapsulation rules, i.e. the interventions should only interact with their targets via the public interface of the target. They shouldn't mess with the private members of the class. Therefore, in order to zero out the outputs of the MLPs, we'll need to add some public property/method to `MirrorModel` to allow us to do so.

With this, we should be able to, in a config file, do something like
```yaml
model:
  class_path: MultiplierIntervention
  init_args:
    multiplier: 2
    target:
      class_path: PlaceholderModel
```

JansenNye:
Tested by running standard Wikitext/Llama training run with `--intervention: MirrorMLPIntervention`, observing that training loss barely decreases over training steps (since the MLPs are outputting zeros and can't learn). 
Closes #79

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -8,6 +8,7 @@ from lightning.fabric.utilities.warnings import PossibleUserWarning
 
 from mirror.config import init_config
 from mirror.datasets.mirror_dataset import MirrorDataset
+from mirror.interventions.intervention import Intervention
 from mirror.models.mirror_model import MirrorModel
 from mirror.models.model_util import instantiate_model
 from mirror.preprocessors.mirror_preprocessor import MirrorPreprocessor
@@ -23,6 +24,7 @@ import mirror.callbacks
 import mirror.datasets
 import mirror.models
 import mirror.preprocessors
+import mirror.interventions
 
 Subcommand = Literal['fit'] | Literal['test'] | Literal['preprocess']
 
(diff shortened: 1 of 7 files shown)


morrisoncheney (approved):
This seems fine to me
I think it is strange to add the property map_modules for this intervention to every model since this is just an example, but I'll let @Plyb judge whether or not that additional code is unnecessary and deserving of being removed.

Plyb (changes requested):

morrisoncheney (approved):
Lookin good!

Plyb (approved):

PR #229, opened 2026-03-30
192 learning rate schedulers

### LINKED ISSUE #192: learning rate schedulers
**Acceptance Criteria**:
- `fit` accepts a customizable learning rate scheduler
- an example for specifying the lr scheduler is present

**Possible design**:
- `Trainer.fit` accepts a `configure_scheduler: ConfigureScheduler | None` parameter
  - If present, a scheduler is constructed using the optimizer
  - if a scheduler is constructed, `scheduler.step()` is called after `optimizer.step()`
  - `class ConfigureScheduler(Protocol): def __call__(self, optimizer: Optimizer, total_training_steps: int) -> Scheduler: ...`
    - `Optimizer` and `Scheduler` are both the built in torch types.
    - You'll need to implement an example of ConfigureScheduler
- `configure_optimizer` flows from the subcommand `fit`'s parameters.

morrisoncheney:
Tested with the classic good ol trusty somebodytoleanon neverletmedown youknowityouloveit WikitextDataset Llamma Model training runs.
(To test I basically just made sure that they both ended up with different losses. The one with the CosineAnnealingLoss stuff ended with a higher loss.)
Closes #192


src/mirror/models/mirror_llama_model.py
```diff
@@ -21,10 +21,13 @@ class MirrorLlamaModel(
     def __init__(
         self,
         initialization: Literal["3.2-1B", "3.2-1B-Instruct"] | LlamaConfig = "3.2-1B-Instruct",
+        seed: int | None = None,
     ) -> None:
         super().__init__()
         self._preprocessor = MirrorLlamaPreprocessor()
         if isinstance(initialization, LlamaConfig):
+            if seed is not None:
+                torch.manual_seed(seed)
```
Plyb, line 30: Good thought to add this, but I don't think this is the right place for it. I actually think it would make sense to put this as a parameter at the top level of the cli (since different things use randomness, like the batch shuffler). The tricky thing is we need to set the seed before the model is created so that we get the random initialization here. Because of that, we'd need to set the seed before initializing the classes with jsonargparse, which means `seed` would need special treatment. Probably best for a follow up ticket.

src/mirror/models/mirror_llama_model.py
```diff
@@ -21,10 +21,13 @@ class MirrorLlamaModel(
...
             self._hf_model = cast(LlamaForCausalLM, AutoModelForCausalLM.from_config(initialization))
```
Plyb, line 31: unrelated to this ticket, but I just realized we should probably just be doing `LlamaForCausalLM.from_config` here

Plyb (approved):
lgtm. I'll create a couple of tickets based on the things I noticed

PR #234, opened 2026-03-31
Fix CUDA unavailable warning

### LINKED ISSUE #233: False "CUDA unavailable" warning
Currently, when setting `device=cpu` on compute nodes, the terminal will print an error saying that CUDA is unavailable and that CPU will be used instead. This is a false error, and CUDA will actually get used rather than CPU. 

Acceptance criteria:
- This "CUDA unavailable" warning only prints when CUDA is truly unavailable and the CPU must actually be used instead

JansenNye:
Tested by running Wikitext Llama training runs with `device=cuda`, `device=cpu`, and locally. 
Closes #233

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -59,8 +59,8 @@ def main(subcommand: Subcommand):
 
             trainer = trainer.construct_trainer()
 
-            trainer.launch()
             if not (is_login_node() and init.slurm.job_type == "compute"):
+                trainer.launch()
                 model = instantiate_model(model, fabric=trainer.fabric)
 
             if is_login_node() and init.slurm.job_type == "local-download":


morrisoncheney (commented):
I tested and this still works with `job_type: local`.

morrisoncheney (approved):
My name is Mo danger c. and I approve this message.

src/main.py
```diff
@@ -59,8 +59,8 @@ def main(subcommand: Subcommand):
 
             trainer = trainer.construct_trainer()
 
-            trainer.launch()
             if not (is_login_node() and init.slurm.job_type == "compute"):
+                trainer.launch()
```
Plyb, line 63: so, this would technically prevent us doing a local training run on a device that has multiple GPUs. But like, that's probably never going to happen? So I'm approving this.

Plyb (approved):

PR #238, opened 2026-04-02
236 add how to use the mirror pipeline section to onboarding doc

### LINKED ISSUE #236: Add "How to use the MIRROR Pipeline" section to onboarding doc
Acceptance criteria:
- A section is added to `mirror onboarding.md` explaining:
   - How to do a training run using existing classes
   - How to use other models/datasets/etc than the ones that are currently in the repo (i.e. explain that implementing the interfaces is the way to do it)

JansenNye:
Tested using markdown file preview I guess
Closes #236

### DIFF
--- a/docs/mirror onboarding.md
+++ b/docs/mirror onboarding.md
@@ -194,21 +194,32 @@ test_data: # Test set
     ... # etc
 ```
 
-Next, pick your SLURM/job type settings. 
+Next, pick your SLURM job/training run settings. 
 
 ```yaml
 slurm:
   job_type: "compute" # "local" "local-download"
   time: "01:00:00" # Timeout limit 
-  gpus_per_node: <gpu_type>:<num_gpus> # GPUs, low->high capability: p100 | a200 (dw87 cluster) | h200  
-  nodes: null
-  mem_per_cpu: "128G"
-  output: "slurm_logs/%j.out"
-  open_mode: "append"
-  signal: "SIGHUP@90"
-  requeue: true
-  qos: <cluster> # Specific GPU cluster
+  gpus_per_node: <gpu_type>:<num_gpus> # GPUs, low -> high capability: p100 | a100 (dw87 cluster) | h200  
+  nodes: 1 # Number of nodes to allocate
+  mem_per_cpu: "128G" # Memory allocated per CPU core
+  output: "slurm_logs/%j.out" # Path for job stdout log (%j = job ID)
+  open_mode: "append" # Append to log file instead of overwriting
+  signal: "SIGHUP@90" # Send SIGHUP signal 90 seconds before timeout to allow graceful shutdown
+  requeue: true # Automatically requeue job if it is preempted
+  qos: <cluster> # Specific GPU cluster/config, e.g. dw87, test
+
+epochs: 1 # Num of passes through the training dataset
+batch_size: 1 # Num of samples in each batch 
+device: cuda # `cuda` for GPU, `cpu` for CPU
 ```
+
+Now submit the training job by running the following command:
+
+`python src/main.py fit --config <configfilename>.yaml`
+
+Replace `fit` with `preprocess` if you are just trying to do a preprocessing run. It's smart to make a separate config file for preprocessing, which won't need parameters like `model`, `val_data`, `test_data`, etc. That way, instead of constantly editing your main config file, you can just pass in your preprocessing config file for preprocessing runs.
+
 
 ### Pre-Pull Request Checklist
 [specific test runs, formatting checks they must run locally before opening a Pull Request]
--- a/docs/mirror onboarding.md
+++ b/docs/mirror onboarding.md
@@ -228,7 +228,7 @@ To test out submitting a training run, run this command:
 python src/main.py fit --config demo_fit_config.yaml
 ```
 
-You should see the output `Submitted batch job <number>`. If this is your first time submitting a training run, you may be prompted to [log in to huggingface] first to download required resources like the Llama model weights.
+You should see the output `Submitted batch job <number>`. If this is your first time submitting a training run, you may be prompted to [log in to huggingface](#initial-access-setup) first to download required resources like the Llama model weights.
 
 That command uses a demo config YAML file to specify the settings for the training run. To customize a training run, you can either use a config file or pass in each argument and its value through the command line, e.g.:
 
@@ -236,39 +236,37 @@ That command uses a demo config YAML file to specify the settings for the traini
 python src/main.py fit --data.class_path WikitextDataset --data.head 10 --model.class_path MirrorLlamaModel
 ```
 
-It's recommended to use config files rather than passing in arguments through the command line. Thus, this section will outline how to use the MIRROR Pipeline through the lens of creating this config file.
+It's recommended to use config files rather than passing in arguments through the command line. Thus, this section will outline how to use the MIRROR Pipeline by walking you through creating this config file, section by section.
 
-First, select your model:
+#### Selecting your model 
+
+If you want to use Llama:
+
 ```yaml
 model:
-    class_path: MirrorLlamaModel # or MirrorGPTModel, etc
-    init_args: # If using Llama: 
+    class_path: MirrorLlamaModel 
+    init_args: 
         initialization:
             # Pretrained weights:
             <model_spec> # 3.2-1B | 3.2-1B-Instruct
+
             # OR custom config:
             vocab_size: <vocab_size>
             hidden_size: <hidden_size> # etc
+```
 
-        # If using GPT2:
+Note that you'll need to [use an h200 GPU](#setting-up-slurm-jobtraining-run-settings) in order to have enough VRAM to run Llama with pretrained weights, so we usually use a smaller custom config of Llama, like the one defined in [demo_fit_config.yaml](../configs/demo_fit_config.yaml).
+
+To use GPT-2:
+
+```yaml
+model:
+    class_path: MirrorGPTModel 
+    init_args: 
         weights: <weights> # pretrained | random
-    # Etc/as appropriate for other models
 ```
 
-Next, decide how the data will be preprocessed. The model's own preprocessor will be used by default, but it can be overridden (see Mirror Architecture/Preprocessors):
-
-```yaml
-preprocessor:
-    class_path: <class_path> # e.g. MirrorLlamaPreprocessor
-```
-
-To preprocess up front rather than on-the-fly:
-
-```yaml
-do_preprocess: True # False by default
-``` 
-
-Now pick your dataset and how it will be used:
+#### Selecting your dataset 
 
 ```yaml
 data: # Training set 
@@ -280,7 +278,11 @@ data: # Training set
     # OR
     # # start_fraction: 0.0
     # end_fraction: 0.8 # Use the first 80% of the selected dataset/split
-    
+```
+
+#### Setting up val/test data
+
+```yaml
 val_data: # Validation set
   class_path: WikitextDataset
   init_args:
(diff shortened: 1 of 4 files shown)


docs/mirror onboarding.md
```diff
@@ -132,37 +132,97 @@ Vim is the default editor for commit message files (e.g. git merge).
...
+```yaml
+preprocessor:
+    class_path: <class_path> # e.g. MirrorLlamaPreprocessor
+```
+
+To preprocess "lazily"/on-the-fly (slower, saves memory):
+
+```yaml
+do_preprocess: True # False by default
+```
```
morrisoncheney: This is actually not correct, you need `do_preprocess: False` in order to preprocess on the fly.

docs/mirror onboarding.md
```diff
@@ -132,37 +132,97 @@ Vim is the default editor for commit message files (e.g. git merge).
...
+```yaml
+data: # Training set 
+  class_path: WikitextDataset
+  init_args:
+    split: train # Use the dataset's `train` portion
+    skip: <i> # Optionally, skip the first <i> examples from the selected split 
+    head: <j> # Use the first <j> examples (after skipping)
+    # OR
+    # start_fraction: 0.0
+    # end_fraction: 0.8 # Use the first 80% of the selected dataset/split
```
morrisoncheney, line 282: I think the start fraction stuff is only implemented on branch 86 if I'm not mistaken.

docs/mirror onboarding.md
```diff
@@ -132,37 +132,97 @@ Vim is the default editor for commit message files (e.g. git merge).
...
+    split: test # Use the dataset's `test` portion
+    ... # etc
+```
+
+Next, pick your SLURM job/training run settings. 
+
+```yaml
+slurm:
+  job_type: "compute" # "local" "local-download"
+  time: "01:00:00" # Timeout limit
```
morrisoncheney: optional: it could be helpful to mention that the requeue callback will continue the job even if the time limit is hit by creating another slurm job submission. That may not need to go right here though.

morrisoncheney (changes requested):
There were a couple of little things to review.

docs/mirror onboarding.md
JansenNye: Whoops

docs/mirror onboarding.md
JansenNye: Are we not going to be merging that branch? If not, we should probably make that same change on a ticket that will be merged, since that's a really useful feature for text datasets without specific splits

docs/mirror onboarding.md
JansenNye: Good idea

docs/mirror onboarding.md
morrisoncheney: I thought Koby had said we weren't gonna, but I could be wrong. 
In any case I think it would make sense to add these to this doc once they are merged to main, rather than having them in here while they "don't exist" for users to access.

I agree. That definitely is a useful feature and keeps users from having to do the extra work involved in figuring out how many rows are in their dataset and such, definitely seems worth implementing in the pipeline.

morrisoncheney (changes requested):
Just the one comment that I have on there.

morrisoncheney (approved):
Lookin good!

docs/mirror onboarding.md
```diff
@@ -132,37 +132,98 @@ Vim is the default editor for commit message files (e.g. git merge).
     - Useful for preparing data separately before running a training job
     - Requires `--data` and `--preprocessor` to be specified (either in the config file or as command-line arguments)
 
+## How to use the MIRROR Pipeline
+
+It's highly recommended to use YAML config files rather than passing in arguments through the command line. Thus, this section will outline how to use the MIRROR Pipeline through the lens of creating this config file.
```
Plyb: This feels like it's missing context. I'd maybe start with something like just telling them the command to run the pipeline using one of the demo configs. Then you can jump in to telling them about the different parts of the config files and how to edit them. The goal here is for someone to be able to just read the first part of this section and make sure they've got the pipeline set up correctly.

docs/mirror onboarding.md
```diff
@@ -132,37 +132,98 @@ Vim is the default editor for commit message files (e.g. git merge).
...
+        initialization:
+            # Pretrained weights:
+            <model_spec> # 3.2-1B | 3.2-1B-Instruct
+            # OR custom config:
+            vocab_size: <vocab_size>
+            hidden_size: <hidden_size> # etc
+
+        # If using GPT2:
+        weights: <weights> # pretrained | random
+    # Etc/as appropriate for other models
```
Plyb: As someone who has never seen this before, how do I know what is "appropriate" for different models? We should tell them what needs to go into these YAML configs in general (i.e. the constructor parameters and how those are translated by jsonargparse)

docs/mirror onboarding.md
```diff
@@ -132,37 +132,98 @@ Vim is the default editor for commit message files (e.g. git merge).
...
+```yaml
+preprocessor:
+    class_path: <class_path> # e.g. MirrorLlamaPreprocessor
+```
+
+To preprocess up front rather than on-the-fly:
+
+```yaml
+do_preprocess: True # False by default
+```
```
Plyb: Could maybe move this down? As someone just starting, I think this would feel like noise. I'd probably be wondering about datasets at this point

docs/mirror onboarding.md
Plyb: yeah, seems like a useful feature. I was planning on merging them in. No real preference on keeping this as is for now or removing it and re-adding it when the feature is merged.

Plyb (changes requested):
One suggestion for how to structure this: think about what questions a person might have as their getting set up and answer those in order. For example, the first question is probably just "did I install this correctly", so letting them know the easiest way to just do a training run with a demo file. Maybe sometime later I'm wondering how to do validation and test datasets, so have a section titled "How to use Val and Test Sets" or something like that. Frame it all around answering "How do I do X"

In all cases though, keep in mind that someone reading it has no context. They don't know anything about our pipeline.

docs/mirror onboarding.md
```diff
@@ -220,6 +220,127 @@ Vim is the default editor for commit message files (e.g. git merge).
...
+Note that you'll need to [use an h200 GPU](#setting-up-slurm-jobtraining-run-settings) in order to have enough VRAM to run Llama with pretrained weights, so we usually use a smaller custom config of Llama, like the one defined in [demo_fit_config.yaml](../configs/demo_fit_config.yaml).
+
+To use GPT-2:
+
+```yaml
+model:
+    class_path: MirrorGPTModel 
+    init_args: 
+        weights: <weights> # pretrained | random
+```
```
Plyb, line 267: Does this accidentally give the impression that there are only two models to use?

docs/mirror onboarding.md
```diff
@@ -220,6 +220,127 @@ Vim is the default editor for commit message files (e.g. git merge).
...
+        initialization:
+            # Pretrained weights:
+            <model_spec> # 3.2-1B | 3.2-1B-Instruct
+
+            # OR custom config:
+            vocab_size: <vocab_size>
+            hidden_size: <hidden_size> # etc
+```
+
+Note that you'll need to [use an h200 GPU](#setting-up-slurm-jobtraining-run-settings) in order to have enough VRAM to run Llama with pretrained weights, so we usually use a smaller custom config of Llama, like the one defined in [demo_fit_config.yaml](../configs/demo_fit_config.yaml).
```
Plyb: I think h200, h100, a100, or b200s should work. Might want to make this more generic

docs/mirror onboarding.md
```diff
@@ -220,6 +220,127 @@ Vim is the default editor for commit message files (e.g. git merge).
...
+
+You should see the output `Submitted batch job <number>`. If this is your first time submitting a training run, you may be prompted to [log in to huggingface](#initial-access-setup) first to download required resources like the Llama model weights.
+
+That command uses a demo config YAML file to specify the settings for the training run. To customize a training run, you can either use a config file or pass in each argument and its value through the command line, e.g.:
+
+```
+python src/main.py fit --data.class_path WikitextDataset --data.head 10 --model.class_path MirrorLlamaModel
+```
+
+It's recommended to use config files rather than passing in arguments through the command line. Thus, this section will outline how to use the MIRROR Pipeline by walking you through creating this config file, section by section.
```
Plyb: I still think somewhere it would be worth mentioning how the yaml files relate to the constructor arguments. I want people to know that if they want to know how to initialize a model/dataset/etc, they can just go look at the class and see what it takes in.

docs/mirror onboarding.md
```diff
@@ -220,6 +220,127 @@ Vim is the default editor for commit message files (e.g. git merge).
...
+```yaml
+val_data: # Validation set
+  class_path: WikitextDataset
+  init_args:
+    split: validation # Use the dataset's `validation` portion
+    ... # etc
+
+val_check_interval: <k> # Perform validation every <k> epochs
+
+test_data: # Test set
```
Plyb: Might be worth mentioning that the test set will be run at the end of the training run

docs/mirror onboarding.md
```diff
@@ -220,6 +220,127 @@ Vim is the default editor for commit message files (e.g. git merge).
...
+  mem_per_cpu: "128G" # Memory allocated per CPU core
+  output: "slurm_logs/%j.out" # Path for job stdout log (%j = job ID)
+  open_mode: "append" # Append to log file instead of overwriting
+  signal: "SIGHUP@90" # Send SIGHUP signal 90 seconds before timeout to allow graceful shutdown
+  requeue: true # Automatically requeue job if it is preempted
+  qos: <cluster> # Specific GPU cluster/config, e.g. dw87, test
+
+epochs: 1 # Num of passes through the training dataset
+batch_size: 1 # Num of samples in each batch 
+device: cuda # `cuda` for GPU, `cpu` for CPU
```
Plyb, line 340: do we actually have a `device` parameter?

docs/mirror onboarding.md
```diff
@@ -220,6 +220,127 @@ Vim is the default editor for commit message files (e.g. git merge).
...
+
+```yaml
+do_preprocess: True # False by default
+``` 
+
+#### Setting up SLURM job/training run settings
+
+```yaml
+slurm:
+  job_type: "compute" # "local" "local-download"
```
Plyb: maybe worth either describing what these options are here, or linking to somewhere that does

docs/mirror onboarding.md
```diff
@@ -220,6 +220,127 @@ Vim is the default editor for commit message files (e.g. git merge).
...
+  time: "01:00:00" # Timeout limit - note that the requeue callback will continue 
+  # the job if the time limit is reached by submitting another SLURM job
+  gpus_per_node: <gpu_type>:<num_gpus> # GPUs, low -> high capability: p100 | a100 (dw87 cluster) | h200
```
Plyb: The gpu list here isn't exaustive
```suggestion
  gpus_per_node: <gpu_type>:<num_gpus> # GPUs, low -> high capability: p100 | a100 (dw87 cluster) | h200 | etc
```

Plyb (changes requested):

docs/mirror onboarding.md
JansenNye: Yes, we had had that whole issue previously where we were setting `device: cpu` due to the false CUDA unavailable warning

Plyb: ah, I see it now. I got confused because it wasn't an argument of the `fit` subcommand function

Plyb (approved):

PR #239, opened 2026-04-08
237 bug requeue callback doesnt work v2

### LINKED ISSUE #237: bug: requeue callback doesn't work
Acceptance Criteria:
- corrected the way that requeue callback stores a ckpt such that it works with the changes implemented in #35.
- Any other changes to ensure that requeue works have been implemented.

morrisoncheney:
Tested by doing a training run that deliberately needed to be requeued several times (llama wiki text).
Closes #237

### DIFF
--- a/src/mirror/callbacks/requeue_callback.py
+++ b/src/mirror/callbacks/requeue_callback.py
@@ -47,8 +47,6 @@ class RequeueCallback[RawT, ProcessedT, BatchT, ModelOutputT](Callback[RawT, Pro
         rank_zero_log(fabric, f"setting up requeue handler on signal {self.requeue_signal}")
         signal.signal(self.requeue_signal, self._make_requeue_handler(fabric))
 
-        self._load_requeue_checkpoint_if_present(fabric, model, optimizer)
-
     def on_train_batch_end(
         self,
         *,
@@ -67,7 +65,7 @@ class RequeueCallback[RawT, ProcessedT, BatchT, ModelOutputT](Callback[RawT, Pro
                 self._requeue(fabric)
             exit()
 
-    def _load_requeue_checkpoint_if_present(
+    def load_requeue_checkpoint_if_present(
         self, fabric: Fabric, model: MirrorModel[RawT, ProcessedT, BatchT, ModelOutputT], optimizer: Optimizer
     ):
         path = requeue_handoff_path()
--- a/src/mirror/callbacks/requeue_callback.py
+++ b/src/mirror/callbacks/requeue_callback.py
@@ -16,16 +16,16 @@ from mirror.models.mirror_model import MirrorModel
 from mirror.slurm_util import get_job_id
 from mirror.util import is_power_of_ten, mirror_data_path
 
+
 def requeue_handoff_path():
     slurm_job_id = get_job_id()
-    return mirror_data_path / 'requeue_handoffs' / f'handoff-{slurm_job_id}.json'
-
-RequeueHandoff = Dict[Literal['previous_training_run_id'], str]
+    return mirror_data_path / "requeue_handoffs" / f"handoff-{slurm_job_id}.json"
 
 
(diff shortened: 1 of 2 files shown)


src/mirror/callbacks/requeue_callback.py
```diff
@@ -16,118 +16,115 @@
...
+        self.requeue_global_step: int | None = None
 
     def on_fit_start(
-            self,
-            *,
-            fabric: Fabric,
-            model: MirrorModel[RawT, ProcessedT, BatchT, ModelOutputT],
-            optimizer: Optimizer,
-            **kwargs,
+        self,
```
JansenNye, line 40: Total nitpick, but the indentation change is wrong I think, it should still be one indent past the start of the function name

src/mirror/callbacks/requeue_callback.py
```diff
@@ -16,118 +16,115 @@
...
                 self._requeue(fabric)
             exit()
 
-    def _load_requeue_checkpoint_if_present(
-            self, 
-            fabric: Fabric, 
-            model: MirrorModel[RawT, ProcessedT, BatchT, ModelOutputT], 
-            optimizer: Optimizer
+    def load_requeue_checkpoint_if_present(
+        self, fabric: Fabric, model: MirrorModel[RawT, ProcessedT, BatchT, ModelOutputT], optimizer: Optimizer
```
JansenNye, line 69: Same inconsequential nitpick, but I'd leave this how it was with an arg on each line

JansenNye: Also, I think this function signature should still start with an underscore, since it's only being used by Trainer - I think convention in Python is that a leading underscore means it's private, no leading underscore means it's public

JansenNye (approved):

src/mirror/callbacks/requeue_callback.py
Plyb: +1 on the leading underscore thing

src/mirror/callbacks/requeue_callback.py
```diff
@@ -16,118 +16,115 @@
...
-                fabric.load(checkpoint_id.path, {
-                    'model': model,
-                    'optimizer': optimizer,
-                })
+                checkpoint_id = self._requeue_checkpoint_id(handoff["previous_training_run_id"])
+                state = {
+                    "model": model,
+                    "optimizer": optimizer,
+                    "global_step": 0,
+                }
```
Plyb, line 85: To stop this problem from happening in the future, what do you think about creating an explicit `State` class that inherits `TypedDict`. That way, if in the future, we ever add anything else to the state (which I think will be likely), the typechecker can catch that it hasn't been added everywhere. We'd need to replace everywhere we're constructing state dicts with this new `State` class.

src/mirror/trainer.py
```diff
@@ -110,42 +107,57 @@ def fit(
...
+                )
             start_epoch = checkpoint_global_step // n_batches
             start_batch = (checkpoint_global_step % n_batches) + 1
 
+        requeue_cb = next((cb for cb in self.callbacks if isinstance(cb, RequeueCallback)), None)
+        if requeue_cb is not None:
+            requeue_cb.load_requeue_checkpoint_if_present(self.fabric, model, optimizer)
+            if requeue_cb.requeue_global_step is not None:
+                start_epoch = requeue_cb.requeue_global_step // n_batches
+                start_batch = (requeue_cb.requeue_global_step % n_batches) + 1
```
Plyb, line 132: Running out of time for this review, but we should rethink this. The trainer shouldn't have to know anything at all about specific callbacks. Unless we can figure out some way to do this from inside the callback, maybe `RequeueCallback` shouldn't be a callback at all. Let's discuss.

Plyb (commented):
Similar question about the linter and the style changes.

Plyb (changes requested):

PR #240, opened 2026-04-08
Thats all that needed to be done for this ticket.

### LINKED ISSUE #186: Add num_proc to our dataset.map calls
Description:
The hugging face Dataset class's map method has a parameter that we can pass in called `num_proc` which will supposedly let us map using multiple nodes. 

Acceptance Criteria:
- Our dataset.map function calls (in both the fit and preprocess subcommands) are run using num_proc set to the correct number (I believe it is just the number of nodes we are using) and the mapping successfully occurs in a distributed manner across multiple nodes.

morrisoncheney:
Tested by running the preprocess subcommand, and fit (llama wikitext).
Closes #186

### DIFF
--- a/src/mirror/subcommands.py
+++ b/src/mirror/subcommands.py
@@ -1,4 +1,3 @@
-
 from dataclasses import asdict
 from pathlib import Path
 import shlex
@@ -6,6 +5,7 @@ import subprocess
 import sys
 
 from jinja2 import Environment, FileSystemLoader, StrictUndefined
+from mirror import slurm_util
 from mirror.checkpoint_identifier import CheckpointIdentifier
 from mirror.datasets.mirror_dataset import MirrorDataset
 from mirror.models.mirror_model import MirrorModel
@@ -16,19 +16,19 @@ from mirror.util import is_login_node
 
 
 def fit(
-        data: MirrorDataset,
-        model: MirrorModel,
-        trainer: Trainer,
-        preprocessor: MirrorPreprocessor | None = None,
-        checkpoint: CheckpointIdentifier | None = None,
-        slurm: SlurmConfig = SlurmConfig(),
-        epochs: int = 1,
-        batch_size: int = 1,
-        do_preprocess: bool = False,
-        run_config_yaml: str = '',
-        val_data: MirrorDataset | None = None,
-        test_data: MirrorDataset | None = None,
-        val_check_interval: int = 1,
+    data: MirrorDataset,
+    model: MirrorModel,
+    trainer: Trainer,
+    preprocessor: MirrorPreprocessor | None = None,
+    checkpoint: CheckpointIdentifier | None = None,
+    slurm: SlurmConfig = SlurmConfig(),
+    epochs: int = 1,
+    batch_size: int = 1,
+    do_preprocess: bool = False,
+    run_config_yaml: str = "",
+    val_data: MirrorDataset | None = None,
+    test_data: MirrorDataset | None = None,
+    val_check_interval: int = 1,
 ):
     if slurm.job_type == "compute" and is_login_node():
         job_id = _submit_slurm_job(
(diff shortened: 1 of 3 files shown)


src/mirror/subcommands.py
```diff
@@ -16,19 +16,19 @@
...
-        checkpoint: CheckpointIdentifier | None = None,
-        slurm: SlurmConfig = SlurmConfig(),
-        epochs: int = 1,
-        batch_size: int = 1,
-        do_preprocess: bool = False,
-        run_config_yaml: str = '',
-        val_data: MirrorDataset | None = None,
-        test_data: MirrorDataset | None = None,
-        val_check_interval: int = 1,
+    data: MirrorDataset,
```
JansenNye, line 19: Same as the other ticket, I think this indentation change is wrong, but obviously it doesn't matter

src/mirror/subcommands.py
```diff
@@ -54,11 +54,8 @@ def fit(
         val_check_interval,
     )
 
-def preprocess(
-        data: MirrorDataset, 
-        preprocessor: MirrorPreprocessor, 
-        slurm: SlurmConfig = SlurmConfig()
-) -> None:
+
+def preprocess(data: MirrorDataset, preprocessor: MirrorPreprocessor, slurm: SlurmConfig = SlurmConfig()) -> None:
```
JansenNye, line 58: Again, similarly, I'd leave each argument on its own line

src/mirror/trainer.py
```diff
@@ -134,18 +131,26 @@ def fit(
         test_dataloader = None
         if test_dataset is not None:
             test_dataloader = self._make_dataloader(test_dataset, preprocessor, batch_size, do_preprocess)
-        
-        self.fabric.call('on_fit_start', fabric=self.fabric, model=model, optimizer=optimizer, dataset=dataset, 
-            training_run_id=training_run_id, n_batches=n_batches, epochs=epochs, start_epoch=start_epoch, 
-            start_batch=start_batch, run_config_yaml=run_config_yaml)
 
-        for epoch_idx in range(start_epoch, epochs):
+        self.fabric.call(
```
JansenNye, line 135: This is the inverse - I changed these fabric calls in a different ticket to have the function arguments more condensed vertically, so that it wasn't taking up all this visual space specifically in the training loop. Regardless, a better solution eventually would be to figure out how to not have to pass all these arguments in at all and visually simplify that way

JansenNye (approved):
Same as the other ticket, I'm not gonna kick it back over indentation and visual style things like that, but if Koby sends it back anyway, I'd revert those changes

src/mirror/subcommands.py
Plyb: I don't really care which way we go, but more changes=more merge conflicts and more noise in a PR, so in the future I'd say unless we have a good reason, leave style changes like this out.

The exception here is if the linter introduced in the typechecking ticket is picky about this. If it is, go with the linter when editing a file.

src/mirror/subcommands.py
```diff
@@ -68,16 +65,11 @@ def preprocess(
...
-
-def _submit_slurm_job(
-        *, 
-        python_args: list[str], 
-        slurm: SlurmConfig, 
-        num_nodes: int,
-        devices: int
-) -> str:
+
+    data.preprocess(preprocessor.preprocess_example, slurm.nodes or 1)
```
Plyb, line 69: tangential thought: maybe it would be handy to have a couple methods on `SlurmConfig` for getting the nodes/devices, both in the presence and absence of a trainer. That way we don't have to always be special-casing how to deal with the `Optional`ity of nodes/devices.

src/mirror/trainer.py
Plyb: yeah, open to suggestions on this one. I'd love to somehow be able to just pass in some kind of "arguments" thing, but I can't think of how to do that without just moving all of these parameters to the line constructing the arguments object.

Plyb (commented):
Two questions/optional things:
1. There is a lot of just style changes in this PR. Was that caused by the linter? If not, let's try to avoid that in the future, since I anticipate this causing a lot of merge conflicts and noise in the PR. However, if it's the linter, it should just be a one time thing anyway.
2. Can you clarify for me how you tested this? How do we know it was actually run on multiple processes?

morrisoncheney:
I just clicked close and not merge, planning on redoing this with the linter off in order to preserve our code's formatting.

PR #242, opened 2026-04-08
Implement configs folder

### LINKED ISSUE #241: Add configs folder
Because of our growing range of applications for the pipeline, e.g. different subcommands/models/etc, each user's list of config files can grow a bit large to be just thrown in the main directory. 
Acceptance criteria:
- A `MIRROR-Pipeline/configs` folder is created in which to place these config files
- The folder's contents are added to the `.gitignore`, replacing `config*.yaml`
- The folder contains one or more demo config files for users to reference or use
- `main.py` automatically parses config files from this folder - i.e., `--config config.yaml` works with `config.yaml` located in the config folder
- The onboarding documentation is updated as necessary

JansenNye:
Tested by calling the standard Wikitext Llama training run with the config file in the `configs` folder.
Closes #241


.gitignore
```diff
@@ -3,6 +3,9 @@ slurm_logs/
 **/__pycache__/
 .ENV
 config*.yaml
+configs/
```
morrisoncheney, line 6: The ticket said that the configs/ would be replacing the configs*.yaml which doesn't seem to be the case here. Not a problem but just something I noticed.

morrisoncheney (approved):

src/mirror/util.py
```diff
@@ -7,6 +7,23 @@
...
+        result.append(args[i])
+        if args[i] == '--config' and i + 1 < len(args):
+            i += 1
+            path = Path(args[i])
+            if not path.exists() and (_CONFIGS_DIR / path).exists():
+                result.append(str(_CONFIGS_DIR / path))
+            else:
+                result.append(args[i])
+        i += 1
+    return result
```
Plyb, line 19: optional: I think this works, but we could maybe express this in code in a way that reflects the underlying logic more directly. It seems like, at a high level, the thing we want to do here is "output a new args list that is the same as the input, except for the config argument. In that case, if the file path doesn't point to a file directly, but does point to one in the configs directory, replace it with the config directory one."

One way to express that would be to convert the list[str] args into a dict (something like this: https://stackoverflow.com/questions/12010126/how-to-parse-an-arbitrary-option-string-into-a-python-dictionary) then just check and modify the one attribute of the dict and convert it back into the list[str].

.gitignore
```diff
@@ -3,6 +3,9 @@ slurm_logs/
...
 mirror_pipeline.egg-info/
+src/mirror/models/baby_babble/*
 .vscode/
 .claude/
+uv.lock
```
Plyb: the baby_babble and uv.lock thing look like the got added from a different branch.

Plyb (changes requested):
One optional thing + some extraneous changes to the .gitignore that should be removed.

.gitignore
JansenNye: Yeah, I realized that 1) I'd have to keep moving my config files back and forth until this gets merged and 2) once it gets merged, everyone would have to change immediately over to the new folder, so I figured it was better to just leave both ways working.

Plyb (approved):

Issue #172, opened 2026-03-11
Minimal load time for "just submission"


Plyb:
Related to #39

Our imports are very slow, and because of our auto-submit logic, we are importing everything twice (once on the login node, once on the compute node). If we're just going to do a slurm submission, can we avoid importing unnecessary packages?

**Acceptance Criteria**:
- Either:
  - only import packages that are strictly necessary for slurm submission (stuff like jsonargparse, jinja, etc)
    - specifically, avoid long loading time packages like torch
  - explain why this isn't possible

I think the hardest thing here is how to deal with classes like `MirrorDataset` that inherit from torch modules. Is there some way we can tell jsonargparse about their schemas without actually loading torch?

One thing to look into: If we _just_ need an import for type annotations, some packages I've seen do a thing where they define a `TYPE_CHECKING` global that is set to true for the type checker, but false otherwise. Then you can do `if TYPE_CHECKING: import torch` or whatever. However, that still doesn't solve the issue of subclasses of torch classes.

PR #245, opened 2026-04-10
39 speed up imports rd

### LINKED ISSUE #39: speed up imports

JansenNye:
Tested by:
   - Running standard job submissions on this branch vs. on the main branch, and confirming that on this branch they take ~5 seconds less to submit on average
   - Running repeat job submissions in the Jupyter notebook, and confirming that the repeat submissions are significantly quicker 

Closes #39 

[Import Speed R&D Report.pdf](https://github.com/user-attachments/files/26639639/Import.Speed.R.D.pdf)


.github/workflows/overfit-pr.yml
```diff
@@ -37,4 +34,4 @@ jobs:
           MIRROR_DATA_PATH: ${{ github.workspace }}/.mirror_data
           MIRROR_PRINT_STEP_LOSS: "true"
         run: |
-          python scripts/check_overfit.py --config overfit_test_config.yaml
+          uv run python scripts/check_overfit.py --config overfit_test_config.yaml
```
morrisoncheney: Just a clarification: you said in the pdf report that you removed the uv stuff, but you've kept it in the overfit-pr. This was intentional, correct?

src/mirror/trainer_constructor.py
```diff
@@ -1,23 +1,29 @@
...
-            strategy: Strategy = FSDPStrategy(),
+            strategy: Strategy | None = None,
             devices: int = 1,
             num_nodes: int = 1,
             callbacks: list[Callback] = [],
     ) -> None:
-        self.strategy = strategy
+        from lightning.fabric.strategies.fsdp import FSDPStrategy
+
+        self.strategy = strategy if strategy is not None else FSDPStrategy()
```
morrisoncheney: Would we want to do something like this to avoid the potentially unnecessary import?
```
if strategy is None:
   from lightning.fabric.strategies.fsdp import FSDPStrategy
   self.strategy = FSDPStrategy()
else:
   self.strategy = strategy
```

.gitignore
```diff
@@ -1,8 +1,17 @@
-slurm_logs/
+# Python
+__pycache__/
+mirror_pipeline.egg-info/
+
+# Environments
 .env/
-**/__pycache__/
```
morrisoncheney, line 4: Will the pycache folders now be untracked and we should just delete them?

morrisoncheney (changes requested):
We just need Jupyter in the dependencies, I also left some other random comments/questions.

.github/workflows/overfit-pr.yml
JansenNye: No, it was not. Good catch haha

src/mirror/trainer_constructor.py
JansenNye: Good idea, not sure how much time or resources this would save but it can't hurt

.gitignore
JansenNye: That should be handled by line 1: `__pycache__/`, I think this was redundant

morrisoncheney (approved):
The notebook seems like they might be working but they aren't loading on my computer. The connection was successfully made, but I am unable to actually get the ipynb to load so I'll pass this on to Koby to finish reviewing.

morrisoncheney:
One other question @JansenNye, is there a way to stop the notebook from running? Like do I need to turn off the server once I am done with the notebook or anything? That seems like it would be important and worth adding to the documentation.

.github/workflows/overfit-pr.yml
```diff
@@ -12,22 +12,19 @@ jobs:
...
         uses: actions/setup-python@v5
         with:
           python-version: "3.12"
-          cache: pip
 
       - name: Install dependencies
-        run: |
-          python -m pip install --upgrade pip
-          pip install -e .
+        run: pip install -e .
```
Plyb, line 27: Because this and the other workflow were edited, we'll need to just keep an eye on the workflows after this is merged to see if they break. If they do, we'll need to add a ticket to fix them.

docs/mirror onboarding.md
```diff
@@ -220,6 +220,9 @@ Vim is the default editor for commit message files (e.g. git merge).
     - Useful for preparing data separately before running a training job
     - Requires `--data` and `--preprocessor` to be specified (either in the config file or as command-line arguments)
 
+- `python src/launch_jupyter.py`: Set up a Jupyter server on a compute node for submitting SLURM jobs in a Jupyter notebook
+    - Jupyter notebooks allow for significantly decreased startup time on repeat job submissions
```
Plyb: nit here: the way you have the jupyter notebook set up, it's not actually submitting jobs (in the sense of SLURM jobs). It is running multiple training runs on the same SLURM job that hosts the jupyter server.

src/mirror/datasets/mirror_dataset.py
```diff
@@ -1,9 +1,8 @@
-from __future__ import annotations
+from abc import abstractmethod
+from sys import stderr
 from typing import Callable, Sequence, Sized
 from torch.utils.data import Dataset
```
Plyb, line 4: did it turn out that there's not really a way to avoid these kinds of imports? ("these" being subclasses to classes we define)?

Plyb (changes requested):
Mostly just questions

High level thoughts:
-  it looks like there is a couple of places where we still need to do heavy imports when we're just launching the slurm job (importing torch when loading the config or the Dataset class as a subclass for instance). Ideally that wouldn't be the case. If we can't make that work, it is what it is, but I want to make sure we explored that.
- I'll prioritize merging this since it has so many changes
- How do you feel about moving all the imports? Is the added complexity worth it?

Plyb (approved):
I went through and looked at how we're doing the slurm launch step. As far as I can tell, there's nothing there that conceptually depends on pytorch, so I think it *should* be possible in principle to get at least that part working.

That said, that would probably take some pretty significant refactoring, and there's already a ticket for that (#172), so approved!

PR #246, opened 2026-04-12
Fix preprocessing type issue

### LINKED ISSUE #171: preprocessing issue: type of __getitem__ is not aligned with self.ds
Related to #155 and #156

On `MirrorDataset`, `__getitem__` returns a specific row type (depending on the exact dataset). But since #8, we're using `MirrorDataset.ds` (a huggingface dataset) as an input to `preprocess`. This inner `.ds` has no restrictions on the types of its rows, but `preprocess` expects that its rows have the same shape as what is returned by `__getitem__`. This can lead to errors.

**Acceptance Criteria**:
- `preprocess` is changed or replaced to resolve the issue above. Our way of preprocessing should be type safe.

**A possible thought**:
what if we added a function that converts a row from the inner `.ds` into a specific row type (such as a `TextRow`). So it could look something like
```python
def to_row_type(self, ds_row: dict): TextRow
    return TextRow(text=ds_row['text_but_named_differently'])
```
Then, in `preprocess` we could have the mappable function use `to_row_type` on the raw ds row before passing it to `preprocess_function`.

This might even be able to fully replace `__getitem__`?

JansenNye:
Tested by preprocessing using MirrorLlamaPreprocessor and Wikitext.
Closes #171

### DIFF
--- a/src/mirror/datasets/mirror_dataset.py
+++ b/src/mirror/datasets/mirror_dataset.py
@@ -1,5 +1,5 @@
 from __future__ import annotations
-from typing import Callable, Sequence, Sized, cast
+from typing import Callable, Sequence, Sized
 from torch.utils.data import Dataset
 from datasets import Dataset as HFDataset
 from abc import abstractmethod
@@ -12,8 +12,9 @@ class MirrorDataset[RawT](Dataset[RawT], Sized):
     def ds(self) -> HFDataset:
         pass
 
+    @abstractmethod
     def to_row_type(self, ds_row: dict) -> RawT:
-        return cast(RawT, ds_row)
+        pass
 
     def __getitem__(self, index: int) -> RawT:
         return self.to_row_type(self.ds[index])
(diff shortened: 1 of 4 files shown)


morrisoncheney (approved):
Looks good to me

src/mirror/datasets/mirror_dataset.py
```diff
@@ -12,20 +12,21 @@ class MirrorDataset[RawT](Dataset[RawT], Sized):
     def ds(self) -> HFDataset:
         pass
 
-    @abstractmethod
+    def to_row_type(self, ds_row: dict) -> RawT:
+        return cast(RawT, ds_row)
```
Plyb: This `cast` here is unsafe. The danger I foresee is that someone is implementing `MirrorDataset`, and the underlying huggingface ds has different names for the columns than the `RawT` specifies (this actually happens in the Mcqa repo. I have `McqaRow`, which specifies a couple of coumns, but not all hf ds's follow those naming conventions), but because there is already this implementation, the typechecker wouldn't warn me about that. The `cast`s were safe on the existing datasets because we know what concrete hf ds is being used with each of them. We don't know that here.

So I think we should have this be abstract and have each of the subclasses implement it separately.

Plyb (changes requested):

Plyb (approved):
lgtm, but there are merge conflicts

PR #247, opened 2026-04-14
Reimplemented requeue callback as requeue monitor.

### LINKED ISSUE #177: Bug: Requeue is broken
When working on the MCQA experiments, I noticed that the requeue didn't work when one of my jobs ran out of time. Not sure if that was a fluke or not. Let's check to make sure it's at least working in general.

might be caused by #200

### LINKED ISSUE #237: bug: requeue callback doesn't work

morrisoncheney:
This is the latest PR for ticket 237.
Closes #237 
Closes #177


JansenNye (approved):

src/mirror/requeue_monitor.py
```diff
@@ -1,52 +1,51 @@
...
-    def on_fit_start(
+    def setup(
             self,
-            *,
             fabric: Fabric,
             model: MirrorModel[RawT, ProcessedT, BatchT, ModelOutputT],
             optimizer: Optimizer,
-            **kwargs,
-    ):
+    ) -> int | None:
```
Plyb: Optional since it's a style thing: Related to my comment below about state, if we passed in a state (instead of a model and optimizer separately) and returned a state, that read read more clearly to me that we're updating the state, and it would include the new/non-updated global_step already. Under the hood it would still just be mutating the state and returning the same object that was passed in, but to someone familiar with functional paradigms, it makes it clearer that this is an "updating" method.

Other advantages:
- makes it so we don't need the comment (comments run the risk of being code duplication)
- could clean up trainer. We wouldn't need the conditional checking for the `None` on the requeue_global_step. If the step is `0`, that is the same as it being `None`. You could just set `requeue_global_step` to whatever `state['global_step']` is after running the checkpoint loading.

In either case, I want it to be *really* clear that this method is modifying the things that it is being passed in. Mutation is dangerous. In this case, it would be pretty inefficient *not* to do mutation, but we should flag it when we are doing it.

(note: I know the old callback was doing mutation. I didn't love it then, but didn't see a good way around it with it being a callback. Moving out from being a callback frees us up a bit).

src/mirror/requeue_monitor.py
```diff
@@ -55,22 +54,22 @@ def on_train_batch_end(
             model: MirrorModel[RawT, ProcessedT, BatchT, ModelOutputT],
             optimizer: Optimizer,
             training_run_id: str,
-            **kwargs,
+            global_step: int,
```
Plyb: see comment on state below. We could pass in a state here instead of model, optimizer, and global_state separately. Might make trainer messier though?

src/mirror/requeue_monitor.py
```diff
@@ -55,22 +54,22 @@ def on_train_batch_end(
...
-            self, 
-            fabric: Fabric, 
-            model: MirrorModel[RawT, ProcessedT, BatchT, ModelOutputT], 
-            optimizer: Optimizer
-    ):
+            self,
+            fabric: Fabric,
+            model: MirrorModel[RawT, ProcessedT, BatchT, ModelOutputT],
+            optimizer: Optimizer,
+    ) -> int | None:
```
Plyb: see comment on the signature of `setup`

src/mirror/requeue_monitor.py
```diff
@@ -81,38 +80,37 @@ def _load_requeue_checkpoint_if_present(
                 handoff: RequeueHandoff = json.loads(handoff_json)
 
                 checkpoint_id = self._requeue_checkpoint_id(handoff['previous_training_run_id'])
-                fabric.load(checkpoint_id.path, {
+                state = {
                     'model': model,
                     'optimizer': optimizer,
-                })
+                    'global_step': 0,
+                }
```
Plyb: Comment on state: I think it would be worth creating a class (idk what to name it besides `State`) that is a `TypedDict` that contains `model`, `optimizer`, and `global_step`. That way whenever we add new things to the state, the typechecker can catch for us if we haven't added it everywhere a state is being constructed and avoid issues like the bug this ticket is fixing in the future.

src/mirror/requeue_monitor.py
```diff
@@ -81,38 +80,37 @@ def _load_requeue_checkpoint_if_present(
...
     def _save_checkpoint(
-            self, 
-            fabric: Fabric, 
-            model: MirrorModel[RawT, ProcessedT, BatchT, ModelOutputT], 
+            self,
+            fabric: Fabric,
+            model: MirrorModel[RawT, ProcessedT, BatchT, ModelOutputT],
             optimizer: Optimizer,
             training_run_id: str,
+            global_step: int,
```
Plyb: could merge into `state`

src/mirror/requeue_monitor.py
```diff
@@ -155,9 +151,8 @@ def _warn_if_iteration_too_long(self, fabric: Fabric):
...
         self.last_train_batch_end_time = current_time
 
-
     def _make_requeue_handler(self, fabric: Fabric):
-        def requeue_hander(_signum: int, _stack: FrameType | None):
+        def requeue_handler(_signum: int, _stack: FrameType | None):
             rank_zero_log(fabric, f'requeue handler called with {_signum}')
             self.requeue_signal_recieved = True
-        return requeue_hander
+        return requeue_handler
```
Plyb, line 134: oops lol. good catch

src/mirror/requeue_monitor.py
```diff
@@ -1,52 +1,51 @@
...
             optimizer: Optimizer,
-            **kwargs,
-    ):
+    ) -> int | None:
+        """Set up the requeue signal handler and load a requeue checkpoint if one is present.
+
+        Returns the global_step from the requeue checkpoint, or None if no checkpoint was loaded.
+        """
         rank_zero_log(fabric, f'setting up requeue handler on signal {self.requeue_signal}')
         signal.signal(self.requeue_signal, self._make_requeue_handler(fabric))
```
Plyb, line 39: I wonder if it might be worth just doing this stuff in the constructor and getting rid of the `setup` method (just calling `load_requeue_checkpoint_if_present` directly where `setup` is being called now). As it stands, `setup` is doing two kind of unrelated things, violating the single responsibility principle.

src/mirror/trainer.py
```diff
@@ -121,11 +127,14 @@ def fit(
...
                 raise RuntimeError(
-                    "checkpoint_global_step cannot be None. " 
+                    "checkpoint_global_step cannot be None. "
                     f"checkpoint '{checkpoint.checkpoint_name}' gave an invalid global step value."
                     )
             start_epoch = checkpoint_global_step // n_batches
             start_batch = (checkpoint_global_step % n_batches) + 1
+        elif requeue_global_step is not None:
+            start_epoch = requeue_global_step // n_batches
+            start_batch = (requeue_global_step % n_batches) + 1
```
Plyb: This will lead to a bug. If someone loads from a checkpoint and that training run gets requeued, it will always just start with the checkpoint they inputted. The requeue checkpoint should take priority. One way to do this is to call `setup`/`load_checkpoint_if_present` after this whole `if` statement and then set `start_epoch` and `start_batch` after that based on whatever the current `state` is.

Plyb (changes requested):
Any details on how this was tested?

morrisoncheney:
> Any details on how this was tested?

Sorry, I had explained testing on my previous two branches of this, and since this is the third iteration I forgot to explain it. I did like a 6000 or so item training run, and set the time limit to be I think about 4 minutes total, so it had to requeue several times in order to finish.

src/mirror/requeue_monitor.py
morrisoncheney: So, you are saying we should take a function that is breaking the single responsibility principle for doing two things, and add it to another function that is probably already doing at least 1 thing (which would lead to a minimum of three things)? That seems a little counter intuitive to me. Would you mind explaining that to me?

morrisoncheney: Also, as the code stands that would mean changing around more things than I first expected to make sure that the functions taken from setup have access to `fabric` which would mean initializing RequeueMonitor outside of the trainer `__init__` function. Doable but seems a little strange. Lmk if you want me to do that.

morrisoncheney:
```
data:
  class_path: WikitextDataset
  init_args:
    head: 600
do_preprocess: True

model:
  class_path: MirrorLlamaModel
  init_args:
    initialization:
      vocab_size: 128256 # Default: 128256
      hidden_size: 256 # Default: 4096
      intermediate_size: 256 # Default: 11008
      num_hidden_layers: 8 # Default: 32
      num_attention_heads: 8 # Default: 32
      num_key_value_heads: null # Default: None

slurm:
  job_type: compute
  time: "00:05:00"
  nodes: 1
  gpus_per_node: "1"
  mem_per_cpu: "128G"
  output: "slurm_logs/%j.out"
  open_mode: "append"
  signal: "SIGHUP@90"
  requeue: true

epochs: 6

device: cuda

```

Above is the config that I used for testing.

JansenNye:
Merge conflicts

JansenNye:
New merge conflicts, hooray!

src/mirror/requeue_monitor.py
```diff
@@ -81,38 +81,37 @@ def _load_requeue_checkpoint_if_present(
...
 
                 checkpoint_id = self._requeue_checkpoint_id(handoff['previous_training_run_id'])
-                fabric.load(checkpoint_id.path, {
+                state : StateDict = {
                     'model': model,
                     'optimizer': optimizer,
-                })
+                    'global_step': 0,
+                }
+                fabric.load(checkpoint_id.path, cast(dict[str, Module | Optimizer | Any], state))
```
Plyb, line 71:
```suggestion
                fabric.load(checkpoint_id.path, cast(StateDict, state))
```

Does this work? I'm pretty sure `X | Any == Any`, so the current type doesn't tell us much.

src/mirror/requeue_monitor.py
```diff
@@ -55,22 +55,22 @@ def on_train_batch_end(
...
-            self, 
-            fabric: Fabric, 
-            model: MirrorModel[RawT, ProcessedT, BatchT, ModelOutputT], 
-            optimizer: Optimizer
-    ):
+            self,
+            fabric: Fabric,
+            model: MirrorModel[RawT, ProcessedT, BatchT, ModelOutputT],
+            optimizer: Optimizer,
+    ) -> StateDict | None:
```
Plyb: related to: https://github.com/DRAGNLabs/MIRROR-Pipeline/pull/247#discussion_r3089794776

Optional

```suggestion
            state: StateDict,
    ) -> StateDict:
```

in the `except` you could just return the passed in state and not deal with the conditional in trainer. Also would allow `setup` to not return ` | None`

src/mirror/requeue_monitor.py
Plyb: Just checking: did you decide not to go with this?

src/mirror/requeue_monitor.py
Plyb: same here, verifying that you decided not to go with this?

src/mirror/requeue_monitor.py
Plyb: Fair questions haha

1. Why move extra things to the constructor? Wouldn't that just make the constructor do more responsibilities? This comes down to what we mean by responsibilities (which admittedly can be pretty subjective!). Responsibilities can be nested inside one another (this is the core idea of decomposition in SWE). As long as you can describe what a function/class/whatever is doing using a single sentence without the word "and", it's probably following the SRP (single responsibility principle). With my proposed change, the constructor would be doing "everything to get the requeue monitor ready for use". Right now that responsibility is split up between the constructor and `setup`. `setup` is *also* doing an extra thing: getting the correct state loaded based on the requeue checkpoint. Another way to put this is that if I were to categorize the things `setup` is doing, the first two lines are more similar to what's currently in the constructor than they are to the last line. Does that clarify things? Lmk if not, happy to explain further.
2. Not sure I'm seeing the challenge here. Wouldn't you just need to move the constructor logic below the `_make_fabric` line in the trainer constructor?

src/mirror/trainer.py
```diff
@@ -110,27 +113,36 @@ def fit(
...
                     "checkpoint_global_step cannot be None. "
                     f"checkpoint '{checkpoint.checkpoint_name}' gave an invalid global step value."
                     )
-            start_epoch = checkpoint_global_step // n_batches
-            start_batch = (checkpoint_global_step % n_batches) + 1
+            
+        if self.requeue_monitor:
+            result = self.requeue_monitor.setup(self.fabric, state)
+            if result is not None:
+                state = cast(StateDict, result)
```
Plyb: just checking: is the cast here necessary? Idk if pyright is smart enough or not to know that result must be `StateDict` here, not `StateDict | None`. It might be.

(also this could be made unnecessary by having setup/_load_requeue_callback_if_present not ever return None)

src/mirror/trainer.py
```diff
@@ -110,27 +113,36 @@ def fit(
...
-            self.fabric.load(checkpoint.path, state)
+            self.fabric.load(checkpoint.path, cast(dict[str, torch.Module | torch.optim.Optimizer | Any], state))
 
-            checkpoint_global_step = state['global_step']
-            if checkpoint_global_step is None:
+            if state['global_step'] is None:
                 raise RuntimeError(
                     "checkpoint_global_step cannot be None. "
                     f"checkpoint '{checkpoint.checkpoint_name}' gave an invalid global step value."
                     )
```
Plyb, line 136: remind me when this condition would happen? The reason I ask is that `global_step` being optional makes us do checks and casts in a couple of places. It would be a lot nicer if we could always just assume it was an `int`. If we need this check still, one way to deal with this would be to have a special `load_state` method that does this check internally and returns a `StateDict`, but we modify `StateDict` so that its `global_step` is just `int`. If we don't need the check, we can also make the tweak to `StateDict` and simplify some things.

src/mirror/callbacks/checkpoint_callback.py
```diff
@@ -63,9 +66,9 @@ def _save_checkpoint(
...
             global_step: int | None,
     ):
-        state = {
+        state : StateDict = {
             'model': model,
             'optimizer': optimizer,
             'global_step': global_step,
         }
-        fabric.save(checkpoint_identifier.path, state)
+        fabric.save(checkpoint_identifier.path, cast(dict[str, Module | Optimizer | Any], state))
```
Plyb, line 74:
```suggestion
        fabric.save(checkpoint_identifier.path, cast(dict[str, StateDict, state))
```
see below

src/mirror/trainer.py
```diff
@@ -110,27 +113,36 @@ def fit(
...
             # models and optimizers are treated specially: they are populated via their load_state_dict
             # methods internally to fabric.load. Anything else in the state dict is just set in place.
 
-            state = {
-                'model': model,
-                'optimizer': optimizer,
-                'global_step': 0,
-            }
-            self.fabric.load(checkpoint.path, state)
+            self.fabric.load(checkpoint.path, cast(dict[str, torch.Module | torch.optim.Optimizer | Any], state))
```
Plyb: Is the cast here actually necessary?

src/mirror/dict_types.py
```diff
@@ -0,0 +1,14 @@
...
+from torch.optim import Optimizer
+
+class TextRow(TypedDict):
+  text: str
+
+class TextLabelRow(TextRow):
+  label: str
+
+class StateDict(TypedDict):
+  model: Module
```
Plyb:
```suggestion
  model: MirrorModel[RawT, etc...]
```
This would mean `StateDict` becomes generic, which further makes it a good idea to change methods that return a state to take one in (so that those methods can get their type arguments from the state parameter)

Edit: actually the comment about taking one in doesn't really make sense if we keep the parameters on `RequeueMonitor` instead of its methods.

src/mirror/trainer.py
```diff
@@ -54,8 +55,10 @@ def __init__(
         ]
         if os.getenv("MIRROR_PRINT_STEP_LOSS", "").lower() == "true":
             default_callbacks.append(PrintStepCallback())
+
+        self.requeue_monitor: RequeueMonitor | None = None
         if config['environment'] == RuntimeEnvironment.SLURM_COMPUTE:
-            default_callbacks.append(RequeueCallback())
+            self.requeue_monitor = RequeueMonitor()
```
Plyb: This either needs type arguments, or we should move the type parameters to the methods of requeue monitor

Plyb (changes requested):
Sorry, this ended up being kind of a beast, didn't it?

There are some new pyright warnings. Could you resolve those?

JansenNye:
Requeuing works, but it restarts at 0, resulting in an infinite loop
Edit: I was on the wrong branch, oopsie

JansenNye (changes requested):
See above

JansenNye (approved):

src/mirror/callbacks/checkpoint_callback.py
morrisoncheney: Double check that we can't do `cast(StateDict, state)`. Send error message to Koby if it won't let you.

src/mirror/trainer.py
morrisoncheney: Wow, good catch! I thought that cast was required but it actually worked when I deleted it.

src/mirror/trainer.py
morrisoncheney: Yeah, turns out this one is.

src/mirror/trainer.py
morrisoncheney: This condition happens for the end checkpoint of a training run. global step is None so that no training will happen (because it's at the end of a training run, so there is no training to be done).

I'm not sure I understand what you mean about different casts being caused by the variability of global step in state.

src/mirror/trainer.py
morrisoncheney: I unfortunately don't think I understand this comment either... .

src/mirror/dict_types.py
morrisoncheney: Done.
Except for that edit. Unless the edit was done. Not sure abt that.

JansenNye:
Worked for me

src/mirror/requeue_monitor.py
Plyb: bump. Did you make the decision not to do this?

src/mirror/requeue_monitor.py
Plyb: bump. decision to not do this?

src/mirror/requeue_monitor.py
```diff
@@ -1,76 +1,74 @@
...
     def _load_requeue_checkpoint_if_present(
-            self, 
-            fabric: Fabric, 
-            model: MirrorModel[RawT, ProcessedT, BatchT, ModelOutputT], 
-            optimizer: Optimizer
-    ):
+            self,
+            fabric: Fabric,
+            model: MirrorModel[RawT, ProcessedT, BatchT, ModelOutputT],
+            optimizer: Optimizer,
```
Plyb: related to the comment immediately below, but this is another place where we could just be passing in `state`. Right now, you're splitting up state just to basically immediately reconstruct it inside this method.

src/mirror/requeue_monitor.py
Plyb: this does not in fact work

src/mirror/requeue_monitor.py
```diff
@@ -81,38 +79,30 @@ def _load_requeue_checkpoint_if_present(
...
                     'model': model,
                     'optimizer': optimizer,
-                })
+                    'global_step': 0,
+                }
+                fabric.load(checkpoint_id.path, cast(dict[str, Module | Optimizer | Any], state))
+                return state
         except FileNotFoundError:
-            pass
+            return None
```
Plyb: If we passed in state to this function, we could return state here, making this function return a non-optional, would would simplify `fit`. Related to the above (Optional).

src/mirror/trainer.py
```diff
@@ -114,27 +117,36 @@ def fit(
...
                     "checkpoint_global_step cannot be None. "
                     f"checkpoint '{checkpoint.checkpoint_name}' gave an invalid global step value."
                     )
-            start_epoch = checkpoint_global_step // n_batches
-            start_batch = (checkpoint_global_step % n_batches) + 1
+
+        if self.requeue_monitor:
+            result = self.requeue_monitor.setup(self.fabric, state)
+            if result is not None:
+                state = result
```
Plyb: This is the bit of main that can be simplified if we make `setup` return state non-Optionally.

Plyb (changes requested):
Just a couple things that I wasn't sure whether just got missed or if you decided to not do them. Also there seems to be a new type error in trainer? But it doesn't show up on any individual line, so idk what's happening there.

Plyb (approved):

PR #248, opened 2026-04-14
implemented necessary changes for this ticket and tested both preprocess

### LINKED ISSUE #186: Add num_proc to our dataset.map calls

morrisoncheney:
and fit method.
Closes #186


JansenNye (approved):

Plyb (approved):
yeah that's so much easier to read haha

PR #286, opened 2026-04-27
262 ssp internet data

### LINKED ISSUE #262: SSP: Internet Data
scoping: we need to figure out what data exactly this will be/how much data we should be getting. Should this just be something like the Pile? Dilma? Something more custom?

**Acceptance Criteria**:
- All Internet-sourced data for SSP is loadable in the Mirror Pipeline

JansenNye:
Tested by running training runs with FinewebDataset and confirming that the dataset downloads & caches properly
Closes #262

### DIFF
--- a/src/mirror/datasets/fineweb_dataset.py
+++ b/src/mirror/datasets/fineweb_dataset.py
@@ -23,12 +23,6 @@ class FinewebDataset(MirrorDataset[TextRow]):
         skip: int | None = None,
         split: Literal['train'] = 'train',
     ):
-        """
-        Args:
-            head: how many examples to include. None includes the whole split.
-            skip: how many examples to skip from the start.
-            split: which dataset split to use.
-        """
         super().__init__()
 
         self._ds = cast(DatasetDict, load_hf_dataset(
--- a/src/mirror/datasets/fineweb_dataset.py
+++ b/src/mirror/datasets/fineweb_dataset.py
@@ -1,0 +1,65 @@
+from typing import Literal, cast
+
+import numpy as np
+from datasets import Dataset, DatasetDict
+
+from mirror.datasets.dataset_util import load_hf_dataset
+from mirror.datasets.mirror_dataset import MirrorDataset
+from mirror.row_types import TextRow
+
+hf_dataset_path = 'HuggingFaceFW/fineweb-edu'
+hf_dataset_name = 'sample-10BT'
+target_token_count = 2_000_000_000
+
+
+class FinewebDataset(MirrorDataset[TextRow]):
+    @property
+    def ds(self) -> Dataset:
+        return self._ds
+
+    def __init__(
+        self,
+        head: int | None = None,
+        skip: int | None = None,
+        split: Literal['train'] = 'train',
+    ):
+        """
+        Args:
+            head: how many examples to include. None includes the whole split.
+            skip: how many examples to skip from the start.
+            split: which dataset split to use.
+        """
+        super().__init__()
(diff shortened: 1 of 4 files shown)


morrisoncheney (approved):
Works and lgtm.

src/mirror/datasets/fineweb_dataset.py
```diff
@@ -0,0 +1,62 @@
...
+        head: int | None = None,
+        skip: int | None = None,
+        split: Literal['train'] = 'train',
+    ):
+        """
+        Args:
+            head: how many examples to include. None includes the whole split.
+            skip: how many examples to skip from the start.
+            split: which dataset split to use.
+        """
```
Plyb: docstring is redundant, can be removed

src/mirror/datasets/fineweb_dataset.py
```diff
@@ -0,0 +1,62 @@
...
+            self._ds = self._ds.select(range(head))
+
+    def _process(self, ds: DatasetDict | Dataset) -> DatasetDict:
+        assert isinstance(ds, DatasetDict)
+        return DatasetDict({split: self._truncate(d) for split, d in ds.items()})
+
+    def _truncate(self, ds: Dataset) -> Dataset:
+        cumulative = np.cumsum(np.asarray(ds['token_count'], dtype=np.int64))
+        cutoff = int(np.searchsorted(cumulative, target_token_count, side='right')) + 1
+        return ds.select(range(cutoff))
```
Plyb, line 47: Is this actually doing anything for us? It seems like we're still downloading the whole dataset, then restricting it to the first 2M tokens. If that's the case, just using `head` seems like it would be easier...

src/mirror/preprocessors/mirror_gpt_preprocessor.py
```diff
@@ -12,14 +12,20 @@
...
+        self._max_length = max_length
 
     def preprocess_example(self, example: TextRow) -> TokenTensor:
-        ids = self._tokenizer.encode(example['text'], add_special_tokens=True)
+        ids = self._tokenizer.encode(
+            example['text'],
+            add_special_tokens=True,
+            max_length=self._max_length,
+            truncation=self._max_length is not None,
+        )
```
Plyb, line 28: were these causing problems without the max length?

Plyb (changes requested):
some minor things and questions

src/mirror/datasets/fineweb_dataset.py
JansenNye: This makes it so only the 2M are used in the cache - the full 10M are never used after the initial download. We could just use `head`, but it would be slower and this should actually be doing something.

src/mirror/preprocessors/mirror_gpt_preprocessor.py
JansenNye: Yes, the Fineweb dataset's examples can be a lot longer than the datasets we've been using like Wikitext, which created some huge memory issues on the GPUs

Plyb: I see... another reason to improve our preprocessing flow! It would be good if we could split these up into multiple training examples instead of just truncating them. I'll link this example in my thoughts about better preprocessing.

Plyb: further question: what is the difference between max_length and truncation here?

Plyb (approved):

PR #289, opened 2026-04-28
Customizable wandb data

### LINKED ISSUE #162: customizable wandb data
Right now, the wandb run only saves training loss. That's helpful, but it would be even more helpful if a user of the pipeline could save other information.

**Acceptance Criteria**:
- There is some way to specify extra information that gets saved in the wandb run

**Possible Design**:
- Do #90 
- Have an extra member on the model output that is something like "training metrics" that is a simple dict.
- Pass those training metrics into the on_train_batch_end callback and have the wandb callback include that

JansenNye:
Tested by running training run with the implemented GradNormMetrics example and confirming that wandb logs grad norm.
Closes #162

### DIFF
--- a/docs/mirror onboarding/mirror onboarding.md
+++ b/docs/mirror onboarding/mirror onboarding.md
@@ -152,6 +152,12 @@ Interventions are model wrappers — MirrorModels that contain a MirrorModel —
 
 Future model interventions will be implemented in this module.
 
+### Metrics
+
+By default, Wandb (`WandbCallback`) only logs `train/loss`, `val/loss`, and `test/loss`. To log additional per-step values to Wandb, pass an `ExtraMetricsGetter` into `WandbCallback`. An `ExtraMetricsGetter` is an abstract base class with a single method, `get_metrics(self, model: MirrorModel) -> dict`, which is called once per training step; whatever dict it returns is merged into that step's wandb log alongside `train/loss`.
+
+Subclasses live in `mirror/metrics/`. For example, `GradNormMetrics(ExtraMetricsGetter)` reads the model's gradients and returns `{"grad_norm": ...}`; to use it, you would [pass `WandbCallback(extra_metrics_getter=GradNormMetrics())` to the trainer's `callbacks=` list](config-example.md), which will override the default `WandbCallback`.
+
 ## Pipeline Developer Information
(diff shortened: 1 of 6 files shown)


src/mirror/callbacks/wandb_callback.py
```diff
@@ -56,12 +59,18 @@ def on_fit_start(
...
         loss: float,
         **kwargs,
     ):
         if self.run:
             self.step += 1
-            self.run.log({"train/loss": loss}, step=self.step)
+            extra_metrics = (
+                self.extra_metrics_getter.get_metrics(model)
+                if self.extra_metrics_getter
+                else {}
```
morrisoncheney: hmmm, interesting notation. Nice!

docs/mirror onboarding/mirror onboarding.md
morrisoncheney, line 1: Nice work updating the documentation. It's clutch that you remembered to do that.

morrisoncheney (approved):
Looks good!

src/mirror/metrics/grad_norm_metrics.py
```diff
@@ -0,0 +1,11 @@
...
+from mirror.models.mirror_model import MirrorModel
+
+
+class GradNormMetrics(ExtraMetricsGetter):
+    def get_metrics(self, model: MirrorModel) -> dict:
+        total_sq = 0.0
+        for p in model.parameters():
+            if p.grad is not None:
+                total_sq += p.grad.detach().pow(2).sum().item()
+        return {"grad_norm": total_sq ** 0.5}
```
Plyb: Has this been tested on multiple nodes? iirc model.parameters() behaves kind of weirdly when you're doing things like fsdp.

I also just want to check that we're not going to run into any numerical instability from summing up all of the squared parameters like this. One option would be to put it all in a tensor together and use pytorch's built in norm methods.

Plyb (changes requested):
a couple questions

src/mirror/metrics/grad_norm_metrics.py
JansenNye: Ah, yeah, I'll fix these

Plyb (approved):

PR #293, opened 2026-05-01
Implement customizable progress callback

### LINKED ISSUE #174: Customizable Progress Callback
Right now, the progress callback only ever displays loss. It would be nice if the user could specify other information to include.

Related to #162 

**Acceptance Criteria**:
- The user can somehow specify other, arbitrary information to be included in the progress callback's labels

JansenNye:
Tested by running standard training run with `Trainer.extra_metrics_getter.class_path: GradNormMetrics` and verifying that ProgressCallback displayed GradNorm
Closes #174


src/mirror/callbacks/progress_callback.py
```diff
@@ -28,10 +28,11 @@ def on_train_batch_end(
             *,
             fabric: Fabric,
             loss: float,
+            extra_metrics: dict,
             **kwargs,
     ):
         if fabric.is_global_zero and self.progress_bar is not None:
-            self.progress_bar.set_postfix(Loss=f"{loss:.3f}", refresh=False)
+            self.progress_bar.set_postfix(Loss=f"{loss:.3f}", **extra_metrics, refresh=False)
```
morrisoncheney: That's neat. I'd have expected something more complicated here.

morrisoncheney (changes requested):
Everything looks good. The only problem is merge conflicts. I'll hit approve as soon as those are gone.

morrisoncheney (dismissed):
Error when ran.

morrisoncheney (approved):
lgtm (**L**ovin**G** **T**his imple**M**entation.)

src/mirror/trainer_constructor.py
```diff
@@ -11,11 +12,19 @@ def __init__(
             devices: int = 1,
             num_nodes: int = 1,
             callbacks: list[Callback] = [],
+            extra_metrics_getter: ExtraMetricsGetter | None = None,
```
Plyb: good catch

Plyb (changes requested):
Some merge conflicts, plus I have one concern:

When it was just the WandB logging, and combined with the ticket to specify how often we log to wandb, it was feasible to specify a more expensive metric but log it less often. The way this is set up now, we are forced to compute the metric every training step, which could be pretty limiting. There are multiple different ways we could deal with this:
1. add some way change the frequency of the extra metrics directly
2. split it up so that the progress callback and the wandb callback can have different sets of extra metrics with the ability to control their frequency independently
3. Maybe something else I'm not thinking of

I don't think I have a strong opinion on this. Maybe slightly lean towards number 2, but not by much.

src/mirror/callbacks/progress_callback.py
```diff
@@ -27,12 +38,15 @@ def on_train_batch_end(
...
             loss: float,
-            extra_metrics: dict,
             **kwargs,
     ):
+        self.step += 1
+        if self.extra_metrics_getter is not None and self.step % self.extra_metrics_every_n_steps == 0:
+            self.last_extra_metrics = self.extra_metrics_getter.get_metrics(model, fabric)
         if fabric.is_global_zero and self.progress_bar is not None:
-            self.progress_bar.set_postfix(Loss=f"{loss:.3f}", **extra_metrics, refresh=False)
+            self.progress_bar.set_postfix(Loss=f"{loss:.3f}", **self.last_extra_metrics, refresh=False)
```
Plyb: Does tqdm require that we have the same entries every time? Or could we have it just show loss on some steps, and the full extra metrics on others? Not a huge deal either way, but it might be slightly cleaner/less confusing to only log the extra metrics on steps where we calculate them, if that's possible. Optional.

Plyb (commented):
optional comment + some merge conflicts

src/mirror/callbacks/progress_callback.py
JansenNye: It does not require that we have the same entries every time, but it seems to me like it'd be more visually stable and less distracting to just show the same metrics every time, and you probably wouldn't be, say, computing loss so infrequently that displaying the last-known value is worse than displaying nothing at all

src/mirror/callbacks/wandb_callback.py
```diff
@@ -23,6 +23,7 @@ def __init__(
         self,
         extra_metrics_getters: list[ExtraMetricsGetter] = [],
         log_every_n_steps: int = 1,
+        extra_metrics_getter: list[ExtraMetricsGetter] = [],
```
Plyb: wandb callback is now taking in two lists of getters

Plyb (changes requested):

src/mirror/callbacks/wandb_callback.py
JansenNye: I'm a prodigy at resolving merge conflicts apparently

Plyb (approved):

PR #296, opened 2026-05-01
Remove preprocess from MirrorDataset

### LINKED ISSUE #216: Extract `preprocess` from `MirrorDataset`
**Acceptance Criteria**:
- the `preprocess` function currently on `MirrorDataset` is extracted to be an independent function (taking in a `MirrorDataset` in addition to its current parameters`

**Why?**
Inheritance is considered a code smell by some. Ideally our abstract classes should be pure interfaces (no implementations), and in this case, there is nothing preventing us from just having that as a separate function.

JansenNye:
Tested by running standard training run with `do_preprocess: True` and confirming that everything still works
Closes #216

### DIFF
--- a/src/mirror/datasets/mirror_dataset.py
+++ b/src/mirror/datasets/mirror_dataset.py
@@ -5,6 +5,7 @@ from torch.utils.data import Dataset
 from datasets import Dataset as HFDataset
 from mirror.util import _ds_cache_path_context
 
+
 class MirrorDataset[RawT](Dataset[RawT], Sized):
     @property
     @abstractmethod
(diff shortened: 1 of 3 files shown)


src/mirror/subcommands.py
```diff
@@ -8,7 +8,7 @@
 
 from mirror.checkpoint_identifier import CheckpointIdentifier
 from mirror.schedulers.configure_scheduler import ConfigureScheduler
-from mirror.datasets.mirror_dataset import MirrorDataset
+from mirror.datasets.mirror_dataset import MirrorDataset, preprocess as preprocess_dataset
```
morrisoncheney: nit: It looks like every time you import you rename it preprocess_dataset. Why not just rename the function?

morrisoncheney (approved):
Looks good, there is one nit that I think is worth fixing, but doesn't affect functionality, just code cleanliness.

src/mirror/datasets/mirror_dataset.py
```diff
@@ -18,15 +19,19 @@ def to_row_type(self, ds_row: dict) -> RawT:
     def __getitem__(self, index: int) -> RawT:
         return self.to_row_type(self.ds[index])
 
-    def preprocess[ProcessedT](self, preprocessor_function: Callable[[RawT], ProcessedT], num_nodes: int) -> Sequence[ProcessedT]:
 
-        def mappable_preprocessor_function(row: dict) -> dict:
-            return {"input_ids": preprocessor_function(self.to_row_type(row))}
+def preprocess[RawT, ProcessedT](
+    dataset: "MirrorDataset[RawT]",
```
Plyb: Question for my benefit more than anything: What does putting the type in a string like this do? What implications will that have?

src/mirror/trainer.py
```diff
@@ -212,11 +212,11 @@ def _eval_loop(self, model, dataloader) -> float:
 
     def _make_dataloader(self, dataset, preprocessor, batch_size, do_preprocess):
         if do_preprocess:
-            preprocessed = dataset.preprocess(preprocessor.preprocess_example, self.num_nodes)
+            preprocessed = preprocess_dataset(dataset, preprocessor.preprocess_example, self.num_nodes)
         else:
             preprocessed = OnDemandPreprocessedDataset(dataset, preprocessor.preprocess_example)
         dataloader = DataLoader(
-            preprocessed,
+            cast(Dataset, preprocessed),
```
Plyb, line 219: Was this cast actually needed before and we just didn't have it? Trying to understand why we're adding it when it didn't used to be there

Plyb (approved):
Just a couple questions

src/mirror/trainer.py
JansenNye: Yeah, the implied type of `preprocessed` changed, so there was a new Pyright warning on this branch that this cast resolves.

src/mirror/datasets/mirror_dataset.py
JansenNye: It makes it a forward reference so the annotation is stored as a string and isn't evaluated at function definition time, but now that preprocess isn't inside the class anymore, this is unnecessary and doesn't do anything

src/mirror/subcommands.py
JansenNye: Yeah, good catch. It's renamed as an import because `preprocess` is already used by something apparently, but I'll just rename it to `preprocess_dataset` unless Koby has an issue with that

PR #297, opened 2026-05-01
202 multinode speed benchmark

### LINKED ISSUE #202: multinode speed benchmark

Plyb:
Closes #202 

What I did to test: Ran the `benchmark_launch.py` script, then ran `benchmark_visualize.py` when some results were back. The table populates properly (including cells that are missing). The chart is also output properly. Note: At the time of writing, there were no errors in the training portion, although a couple of them did have errors earlier on in the setup (these didn't get caught by the timer callback, but I don't think it's a big enough issue to worry about). I'm not yet sure what the errors were about, they seemed to be somewhat sporadic. Maybe something to do with particular combinations of supercomputer nodes? In any case, the vast majority of cases worked. I didn't test any models/batch sizes large enough to hit an OOM error. I also never ran into a timeout in my testing.


scripts/train-time-benchmark/benchmark_launch.py
```diff
@@ -0,0 +1,160 @@
+#!/usr/bin/env python
+"""Grid-search benchmark launcher (runs on login node).
+
+For each combination of device configuration, batch size, and model size,
+submits an sbatch job unless the configuration is already running (lock file
+exists) or already completed (entry in the JSONL log).
+"""
```
Plyb, line 7: I decided to leave these docstrings at the beginning of the scripts. Because I don't anticipate us editing scripts much, there is less risk of desync between the code and comments.

scripts/train-time-benchmark/benchmark_launch.py
```diff
@@ -0,0 +1,160 @@
...
+from mirror.util import count_params, mirror_data_path
+
+
+DEVICE_CONFIGS: list[tuple[int, int]] = [
+    (1, 1),
+    (1, 8),
+    (2, 8),
+    (4, 8),
+]
+BATCH_SIZES: list[int] = [1, 4, 16, 64]
```
Plyb, line 27: my thought here is that people wanting to use this script can just tweak these directly as they see fit.

src/mirror/benchmark_configs.py
```diff
@@ -0,0 +1,8 @@
+from mirror.models.configuration_llama import LlamaConfig
+
+MODEL_CONFIGS: dict[str, LlamaConfig] = {
+    "100k": LlamaConfig(vocab_size=6000, hidden_size=16,  intermediate_size=24,   num_hidden_layers=4,  num_attention_heads=16, tie_word_embeddings=True),
+    "1M":   LlamaConfig(vocab_size=8000, hidden_size=96,  intermediate_size=128,  num_hidden_layers=4,  num_attention_heads=16, tie_word_embeddings=True),
+    "10M":  LlamaConfig(vocab_size=8000, hidden_size=360, intermediate_size=400,  num_hidden_layers=8,  num_attention_heads=16, tie_word_embeddings=True),
+    "100M": LlamaConfig(vocab_size=8000, hidden_size=896, intermediate_size=1000, num_hidden_layers=16, num_attention_heads=16, tie_word_embeddings=True),
+}
```
Plyb, line 8: Somewhat annoying that this has to be in the main project. The issue is that scripts can't easily import stuff from one another, and this is used both in the run and launch scripts.

scripts/train-time-benchmark/benchmark_launch.py
```diff
@@ -0,0 +1,160 @@
...
+        #SBATCH --qos={qos}
+        #SBATCH --output=slurm_logs/%j.out
+        #SBATCH --open-mode=append
+        #SBATCH --signal=SIGHUP@90
+        #SBATCH --chdir={Path.cwd()}
+
+        mamba activate ./.env
+
+        {run_cmd}
+    """)
```
Plyb, line 95: Why not just run `fit` and let its launcher logic handle this? `fit` uses the exact flags passed to whatever script it's running in. We need to pass different flags in this case.

src/mirror/preprocessors/bpe_preprocessor.py
```diff
@@ -0,0 +1,62 @@
+import hashlib
```
Plyb, line 1: Thanks for letting me steal this

JansenNye:
lgtm + works (when I run the right file!)

src/mirror/preprocessors/bpe_preprocessor.py
morrisoncheney: Hehehe. Tech Lead who??

pyproject.toml
```diff
@@ -16,6 +16,7 @@ dependencies = [
     "lightning==2.5.5",
     "transformers==5.2.0",
     "tqdm==4.67.1",
+    "matplotlib==3.10.9",
```
morrisoncheney, line 19: 'bout time we got this in here!

morrisoncheney (approved):
Honestly, it works.. and Koby wrote it lol. It looks good but I don't think it's worth me digging too deeply into it.

PR #298, opened 2026-05-02
Successfully refactor imports to cut submission time

### LINKED ISSUE #172: Minimal load time for "just submission"

JansenNye:
Tested by:
 - Submitting a standard training run to SLURM and observing that submission time took less than a second, down from ~60 seconds
 - Testing a local training run to confirm that that still works too
 
Closes #172

### DIFF
--- a/src/mirror/slurm_launcher.py
+++ b/src/mirror/slurm_launcher.py
@@ -1,0 +1,89 @@
+import shlex
+import socket
+import subprocess
+import sys
+from dataclasses import asdict, fields
+from pathlib import Path
+
+import yaml
+from jinja2 import Environment, FileSystemLoader, StrictUndefined
+
+from mirror.slurm_util import SlurmConfig
+from mirror.util import resolve_config_args
+
+
+def submit_slurm_job(python_args: list[str]) -> str | None:
+    """Submit a slurm job if `slurm.job_type` is 'compute' on a login node.
+
+    `python_args` is everything after the python script (e.g. `sys.argv[1:]`),
+    starting with the subcommand. Returns the slurm job id if a job was
+    submitted, otherwise None — the caller should bail out only on a non-None
+    return.
+    """
+    if 'login' not in socket.gethostname():
+        return None
+
+    sections = _collect_sections(resolve_config_args(python_args), ('slurm', 'trainer'))
+    valid = {f.name for f in fields(SlurmConfig)}
+    slurm = SlurmConfig(**{k: v for k, v in sections['slurm'].items() if k in valid})
+
+    if slurm.job_type != "compute":
+        return None
+
+    trainer = sections['trainer']
+    ctx = asdict(slurm)
+    ctx["nodes"] = ctx["nodes"] or trainer.get('num_nodes') or 1
+    ctx["ntasks_per_node"] = ctx["ntasks_per_node"] or trainer.get('devices') or 1
+    ctx["gpus_per_node"] = ctx["gpus_per_node"] or trainer.get('devices') or 1
+
+    env = Environment(
+        loader=FileSystemLoader(Path(__file__).parent / "templates"),
+        undefined=StrictUndefined,
+        trim_blocks=True,
+        lstrip_blocks=True,
+    )
+    script = env.get_template("slurm.jinja").render(
+        **ctx,
+        chdir=str(Path.cwd()),
+        activate_cmd="mamba activate ./.env",
+        run_cmd=f"srun python {sys.argv[0]} {shlex.join(python_args)}",
+    )
+
+    res = subprocess.run(["sbatch"], input=script, text=True, capture_output=True)
+    if res.returncode != 0:
+        raise RuntimeError(
+            f"sbatch failed (exit {res.returncode}):\n{res.stderr}\n\nGenerated script:\n{script}"
+        )
+    job_id = res.stdout.strip().split()[-1]
+    print(f"Submitted batch job {job_id}")
+    return job_id
+
+
+def _collect_sections(args: list[str], names: tuple[str, ...]) -> dict[str, dict]:
+    """Collect the named top-level keys from --config files and --<name>.<key>[=value] CLI args."""
+    out: dict[str, dict] = {n: {} for n in names}
+    i = 0
+    while i < len(args):
+        a = args[i]
+        if a == '--config' and i + 1 < len(args):
+            with open(args[i + 1]) as f:
+                data = yaml.safe_load(f) or {}
+            for n in names:
+                out[n].update(data.get(n) or {})
+            i += 2
+            continue
+        section = next((n for n in names if a.startswith(f'--{n}.')), None)
+        if section is None:
+            i += 1
+            continue
+        key, eq, value = a[len(section) + 3:].partition('=')
+        if eq:
+            i += 1
+        elif i + 1 < len(args) and not args[i + 1].startswith('--'):
+            value = args[i + 1]
+            i += 2
+        else:
+            i += 1
+            continue
+        out[section][key] = yaml.safe_load(value)
+    return out
--- a/src/mirror/slurm_launcher.py
+++ b/src/mirror/slurm_launcher.py
@@ -1,40 +1,25 @@
 import shlex
-import socket
 import subprocess
 import sys
-from dataclasses import asdict, fields
+from dataclasses import asdict
 from pathlib import Path
 
-import yaml
 from jinja2 import Environment, FileSystemLoader, StrictUndefined
 
 from mirror.slurm_util import SlurmConfig
-from mirror.util import resolve_config_args
+from mirror.util import is_login_node
 
 
-def submit_slurm_job(python_args: list[str]) -> str | None:
-    """Submit a slurm job if `slurm.job_type` is 'compute' on a login node.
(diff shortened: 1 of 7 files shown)


morrisoncheney (approved):
Worked much faster, and I tested a local run to make sure that still worked and it did.

src/mirror/slurm_launcher.py
```diff
@@ -0,0 +1,89 @@
...
+
+def submit_slurm_job(python_args: list[str]) -> str | None:
+    """Submit a slurm job if `slurm.job_type` is 'compute' on a login node.
+
+    `python_args` is everything after the python script (e.g. `sys.argv[1:]`),
+    starting with the subcommand. Returns the slurm job id if a job was
+    submitted, otherwise None — the caller should bail out only on a non-None
+    return.
+    """
+    if 'login' not in socket.gethostname():
```
Plyb: I thought we had an `is_login_node` function somewhere? Or maybe it was just `is_compute_node`? If not, optional: add it. If we do have it, let's use it here.

EDIT: no yeah, it definitely exists

src/mirror/slurm_launcher.py
```diff
@@ -0,0 +1,89 @@
...
+        return None
+
+    sections = _collect_sections(resolve_config_args(python_args), ('slurm', 'trainer'))
+    valid = {f.name for f in fields(SlurmConfig)}
+    slurm = SlurmConfig(**{k: v for k, v in sections['slurm'].items() if k in valid})
```
Plyb: optional: extract to a "parse arguments" function

(on a more general note, this looks like it might be agent generated? For the style I like personally, agents tend to write functions that are too long/not atomic enough. I either have to explicitly tell it in my agent instructions to write short function (no more than 10 lines is a good heuristic) or correct it myself when using an agent. Other people have different styles they like though)

Plyb: Is the reason we're explicitly constructing the `SlurmConfig` here to make sure we're not missing any members?

src/mirror/slurm_launcher.py
```diff
@@ -0,0 +1,89 @@
...
+
+    if slurm.job_type != "compute":
+        return None
+
+    trainer = sections['trainer']
+    ctx = asdict(slurm)
+    ctx["nodes"] = ctx["nodes"] or trainer.get('num_nodes') or 1
+    ctx["ntasks_per_node"] = ctx["ntasks_per_node"] or trainer.get('devices') or 1
+    ctx["gpus_per_node"] = ctx["gpus_per_node"] or trainer.get('devices') or 1
```
Plyb: optional: extract to "build jinja context" function

Plyb: I'm worried about code duplication with this defaulting behavior. Worth thinking about if there is a way we can consolidate that.

src/mirror/slurm_launcher.py
```diff
@@ -0,0 +1,89 @@
...
+    if res.returncode != 0:
+        raise RuntimeError(
+            f"sbatch failed (exit {res.returncode}):\n{res.stderr}\n\nGenerated script:\n{script}"
+        )
+    job_id = res.stdout.strip().split()[-1]
+    print(f"Submitted batch job {job_id}")
+    return job_id
+
+
+def _collect_sections(args: list[str], names: tuple[str, ...]) -> dict[str, dict]:
```
Plyb: The return type on this isn't correct. If anything, it should be `dict[str, object]` or `dict[str, Any]`. I'm wondering if there's some more type-safe way to do this though (considering we're assuming things about the return values).

My first thought is that we have it do one name at a time and pass it a TypedDict constructor (such as SlurmConfig).
Something like
```
def test[T](cons: type[T]) -> T:
    return cons()
```
but have it get the arguments for the constructor from the python arguments

src/mirror/slurm_launcher.py
```diff
@@ -0,0 +1,89 @@
...
+    """Collect the named top-level keys from --config files and --<name>.<key>[=value] CLI args."""
+    out: dict[str, dict] = {n: {} for n in names}
+    i = 0
+    while i < len(args):
+        a = args[i]
+        if a == '--config' and i + 1 < len(args):
+            with open(args[i + 1]) as f:
+                data = yaml.safe_load(f) or {}
```
Plyb: minor thing, not sure we want to worry about this: jsonargparse does allow json as well as yaml config files. This would restrict us to yaml. Probably fine though...

src/mirror/slurm_launcher.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+def submit_slurm_job(python_args: list[str]) -> str | None:
+    """Submit a slurm job if `slurm.job_type` is 'compute' on a login node.
+
+    `python_args` is everything after the python script (e.g. `sys.argv[1:]`),
+    starting with the subcommand. Returns the slurm job id if a job was
+    submitted, otherwise None — the caller should bail out only on a non-None
+    return.
+    """
```
Plyb, line 20: The way this is structured, it's not currently super reusable for other projects. It assumes that slurm and trainer are in the python args, which may not be the case for projects importing the mirror pipeline.

My recommendation:
- split out the `SlurmConfig` parsing code into its own function, and have this take in a `SlurmConfig` instead.
  - call the slurm config parser in main. Then, if other projects need to parse the python args, they can, otherwise they can just construct the slurm config however is appropriate for their project
- Have this function do the aborting if it submits to slurm. That way the caller doesn't have to worry about it
  - This means this function can just return `None | Never`.

Plyb:
Also some merge conflicts

JansenNye:
Re requesting tech lead review cuz resolving some of the merge conflicts required a bit of chicanery and shenanigans

src/mirror/util.py
```diff
@@ -41,6 +45,10 @@ def get_device() -> str:
 def is_power_of_ten(n: int):
     return n > 0 and math.log10(n).is_integer()
 
+def count_params(model: "nn.Module") -> int:
```
Plyb, line 48: Does this not work without putting it in quotes?

src/mirror/util.py
```diff
@@ -3,6 +3,10 @@
 from collections.abc import Generator
 from contextlib import contextmanager
 from pathlib import Path
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    import torch.nn as nn
```
Plyb, line 9: optional: it may be worth just pulling the count_params function into a new file, like `torch_utils` or something.

Plyb (changes requested):
oops, I apparently didn't submit my review last time!

src/mirror/slurm_launcher.py
JansenNye: Good idea and good note

JansenNye: No, cuz SlurmConfig has defaults anyway; it's to filter out unknown keys ahead of time, but that seems pretty unnecessary so I just got rid of it

src/mirror/slurm_launcher.py
JansenNye: I can't imagine how we'd get it consolidated below three lines, and this is more readable than mapping + a loop to me

src/mirror/slurm_launcher.py
JansenNye: Are you thinking of the contents of the second `dict`? We could definitely change that to `dict[str, dict[str, Any]]` whereas right now it's implicitly `dict[str, dict[Any, Any]]`. I don't think the return type should be `dict[str, object]` or `[dict[str, Any]]` since `dict[str, dict]]` accurately describes how each section name maps to a dict.

src/mirror/util.py
JansenNye: Yeah, since `nn` only exists when type checking

src/mirror/util.py
JansenNye: I don't think this is worth it for just one function/2 lines, but worth keeping in mind for the future

src/mirror/slurm_launcher.py
Plyb: oh, sorry, I meant that I thought there were a few other places in the code we were doing this defaulting behavior between the trainer and the slurm config, but maybe not. This section on its own is totally fine.

src/mirror/slurm_launcher.py
Plyb: ah, I had misunderstood what this function was doing. Carry on.

Plyb (approved):

Plyb (approved):

PR #304, opened 2026-05-06
217 batched flatmap preprocessing

### LINKED ISSUE #217: Batched (flatmap) preprocessing
Using batched `map` on a huggingface dataset allows a single row of the input dataset to become an arbitrary number of rows in the output dataset. This is useful for something like the MCQA experiments, since a single multiple choice question could become multiple training examples (one for each option in the mcqa problem).

**Acceptance Criteria**:
- the `preprocess` function for `MirrorDataset` uses `batched=True`
- The types and return values are updated: preprocessors should now return a `Sequence[OutT]` (might need to be a `List` specifically? not sure) instead of just `OutT`.

JansenNye:
Tested using this Claude script:
```
cd /home/jansen05/MIRROR-Pipeline

PYTHONPATH=src .env/bin/python <<'EOF'
from datasets import Dataset as HFDataset
from mirror.datasets.mirror_dataset import preprocess_dataset
from mirror.datasets.on_demand_preprocessed_dataset import OnDemandPreprocessedDataset

class FakeDS:
    def __init__(self, ds): self._ds = ds
    @property
    def ds(self): return self._ds
    def to_row_type(self, row): return row['text']
    def __len__(self): return len(self._ds)
    def __getitem__(self, i): return self.to_row_type(self._ds[i])

raw = HFDataset.from_dict({'text': ['hello world', 'foo bar', 'baz']})
ds = FakeDS(raw)

one_to_one = lambda t: [list(t.encode())]
flatmap    = lambda t: [[1, len(t)], [2, len(t)]]

print('--- preprocess_dataset ---')
print('1:1 len:', len(preprocess_dataset(ds, one_to_one, num_nodes=1)), '(expected 3)')
out = preprocess_dataset(ds, flatmap, num_nodes=1)
print('flatmap len:', len(out), '(expected 6)')
print('flatmap [0]:', out[0].tolist(), '  [1]:', out[1].tolist())

print('--- OnDemandPreprocessedDataset ---')
od1 = OnDemandPreprocessedDataset(ds, one_to_one)
print('1:1 len:', len(od1), '  sample:', od1[0])
odf = OnDemandPreprocessedDataset(ds, flatmap)
print('flatmap len:', len(odf), '  [5]:', odf[5], '(expected [2, 3])')
EOF
```
And confirming that the output was:
```
1:1 len: 3, flatmap len: 6, OD flatmap len: 6, [5]: [2, 3]
```

Closes #217

### DIFF
--- a/src/mirror/datasets/on_demand_preprocessed_dataset.py
+++ b/src/mirror/datasets/on_demand_preprocessed_dataset.py
@@ -14,7 +14,6 @@ class OnDemandPreprocessedDataset[RawT, ProcessedT](Dataset[ProcessedT]):
         super().__init__()
         self.raw_dataset = raw_dataset
         self.preprocess = preprocess
-        # cumulative_lengths[i] = total processed items produced by raw_dataset[:i+1]
         cumulative = 0
         self._cumulative_lengths: list[int] = []
         for i in range(len(raw_dataset)):
--- a/src/mirror/datasets/on_demand_preprocessed_dataset.py
+++ b/src/mirror/datasets/on_demand_preprocessed_dataset.py
@@ -1,22 +1,31 @@
 from __future__ import annotations
-from typing import Callable
+from bisect import bisect_right
+from typing import Callable, Sequence
 from torch.utils.data import Dataset
 
 from mirror.datasets.mirror_dataset import MirrorDataset
 
 class OnDemandPreprocessedDataset[RawT, ProcessedT](Dataset[ProcessedT]):
     def __init__(
-            self, 
-            raw_dataset: MirrorDataset[RawT], 
-            preprocess: Callable[[RawT], ProcessedT]
+            self,
+            raw_dataset: MirrorDataset[RawT],
+            preprocess: Callable[[RawT], Sequence[ProcessedT]]
     ):
         super().__init__()
         self.raw_dataset = raw_dataset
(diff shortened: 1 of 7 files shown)


morrisoncheney (approved):
Lgtm. (lovely great totally money.)

src/mirror/datasets/on_demand_preprocessed_dataset.py
```diff
@@ -1,22 +1,30 @@
...
+            preprocess: Callable[[RawT], Sequence[ProcessedT]]
     ):
         super().__init__()
         self.raw_dataset = raw_dataset
         self.preprocess = preprocess
+        cumulative = 0
+        self._cumulative_lengths: list[int] = []
+        for i in range(len(raw_dataset)):
+            cumulative += len(preprocess(raw_dataset[i]))
+            self._cumulative_lengths.append(cumulative)
```
Plyb, line 21: This breaks the on demand nature of this wrapper in a pretty fundamental way. We'll need to rethink this ticket I think. Maybe we need to be okay with iterable datasets that don't have a length? But it would still be nice to have a progress bar. Maybe the progress bar is somehow tied to the original dataset instead of the processed one?

Could I have you look into this and figure out what our best options are?

Plyb (changes requested):

PR #308, opened 2026-05-07
Extract Dataloader construction method

### LINKED ISSUE #214: Extract Dataloader construction
I'd like to be able to conveniently set up a dataloader in the same way the `Trainer` does for other purposes, so let's pull it out of the `Trainer`.

**Acceptance Criteria**:
- `Trainer._make_dataloader` is decoupled from `Trainer` (i.e. it is its own separate function)
  - Can take in an optional `Fabric`. If `Fabric` is not given, it doesn't do `setup_dataloaders`
  - `config` can be replaced with `get_config()`

JansenNye:
```
import torch
from datasets import Dataset as HFDataset
from lightning import Fabric
from lightning.fabric.strategies.single_device import SingleDeviceStrategy
from torch.utils.data import DataLoader

from mirror.config import init_config
from mirror.datasets.mirror_dataset import MirrorDataset
from mirror.trainer import make_dataloader


class TinyDataset(MirrorDataset[int]):
    def __init__(self, values: list[int]) -> None:
        self._ds = HFDataset.from_list([{"x": v} for v in values])

    @property
    def ds(self) -> HFDataset:
        return self._ds

    def to_row_type(self, ds_row: dict) -> int:
        return ds_row["x"]

    def __len__(self) -> int:
        return len(self._ds)


class TinyPreprocessor(MirrorPreprocessor[int, torch.Tensor, torch.Tensor]):
    def preprocess_example(self, example: int) -> torch.Tensor:
        return torch.tensor([example])

    def collate(self, examples: list[torch.Tensor]) -> torch.Tensor:
        return torch.stack(examples)


def test_without_fabric_returns_plain_dataloader() -> None:
    dataset = TinyDataset([1, 2, 3, 4])
    preprocessor = TinyPreprocessor()
    dl = make_dataloader(dataset, preprocessor, batch_size=2,
                         do_preprocess=False, shuffle=False)
    assert isinstance(dl, DataLoader)
    batches = list(dl)
    assert len(batches) == 2
    assert torch.equal(batches[0], torch.tensor([[1], [2]]))
    assert torch.equal(batches[1], torch.tensor([[3], [4]]))


def test_with_fabric_wraps_dataloader() -> None:
    dataset = TinyDataset([10, 20, 30, 40])
    preprocessor = TinyPreprocessor()
    fabric = Fabric(strategy=SingleDeviceStrategy(device="cpu"), accelerator="cpu", devices=1)
    fabric.launch()
    dl = make_dataloader(dataset, preprocessor, batch_size=2,
                         do_preprocess=False, shuffle=False, fabric=fabric)
    plain = make_dataloader(dataset, preprocessor, batch_size=2,
                            do_preprocess=False, shuffle=False)
    assert type(dl) is not type(plain)
    batches = list(dl)
    assert len(batches) == 2
    assert torch.equal(batches[0], torch.tensor([[10], [20]]))


if __name__ == "__main__":
    init_config(device="cpu")
    test_without_fabric_returns_plain_dataloader()
    print("PASS: without fabric returns plain DataLoader")
    test_with_fabric_wraps_dataloader()
    print("PASS: with fabric returns wrapped DataLoader")
```
Closes #214

### DIFF
--- a/src/mirror/trainer.py
+++ b/src/mirror/trainer.py
@@ -107,7 +107,10 @@ class Trainer[RawT, ProcessedT, BatchT, ModelOutputT]:
         )
 
 
-        dataloader = self._make_dataloader(dataset, preprocessor, batch_size, do_preprocess, shuffle)
+        dataloader = make_dataloader(
+            dataset, preprocessor, batch_size, do_preprocess, shuffle,
+            num_nodes=self.num_nodes, fabric=self.fabric,
+        )
 
         start_epoch = 0
         start_batch = 0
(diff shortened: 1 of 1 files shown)


morrisoncheney (changes requested):
Merge conflicts.

morrisoncheney (approved):
lgtm. (lovely-goodness-to-my-code-senses.)

Plyb (approved):

PR #317, opened 2026-05-10
Implement submission time PR test

### LINKED ISSUE #307: Add SLURM submission time PR test
Acceptance criteria:
 - If it's possible to do so, here is a third GitHub Actions PR test that checks that SLURM submission time remains low/near instant, protecting against accidentally blowing it back up with a new import somewhere
 - If this is not possible (e.g. you can't submit to the supercomputer through GitHub Actions), any potential alternative solutions to checking/testing this are brainstormed
   - If an alternative solution is reasonably simple, it is implemented
   - If an alternative solution is more complex, a new ticket is created to implement it
   - If an alternative solution does not exist or is not reasonable, a brief report is created explaining why

JansenNye:
Testing by making this pull request (and tested by Claude through various custom scripts it ran)
Closes #307


morrisoncheney (approved):
lgtm (looks-good-to-myclaudetoo)

.github/workflows/submission-time-pr.yml
```diff
@@ -0,0 +1,30 @@
...
+      github.event.pull_request.author_association == 'MEMBER' ||
+      github.event.pull_request.author_association == 'COLLABORATOR' ||
+      github.event.pull_request.author_association == 'CONTRIBUTOR'
+
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v4
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
```
Plyb: weren't these version the ones causing the deprecation warning from the other ticket?

scripts/check_submission_time.py
```diff
@@ -0,0 +1,76 @@
...
+def _invoke_main(config: str) -> None:
+    sys.argv = [str(_REPO_ROOT / "src/main.py"), "fit", "--config", config]
+    sys.path.insert(0, str(_REPO_ROOT / "src"))
+    import main as main_module
+    main_module.main("fit")
+
+
+def main() -> int:
+    args = _parse_args()
+    recorded: list[float] = []
```
Plyb: using a list here is slightly confusing imo since we're never going to include more than one entry. maybe a tiny dataclass that just has a float member would be better?

scripts/check_submission_time.py
```diff
@@ -0,0 +1,76 @@
...
+    start = time.perf_counter()
+    _patch_login_node()
+    _patch_sbatch(start, recorded)
```
Plyb: Optional: the elapsed time is going to include the patching. This is incredibly minimal, so it probably doesn't matter, but if we do want to be more precise, we could have `recorded` just contain the end time, and do `start` right before the `invoke_main` call.

Plyb (changes requested):
Seems like a good setup overall, just a couple comments

.github/workflows/submission-time-pr.yml
JansenNye: Oh yeah haha I might have done this before I fixed that

Plyb (approved):

PR #319, opened 2026-05-12
303 create metric architecture and fluency metric

### LINKED ISSUE #303: Create Metric Architecture and Fluency Metric
**Acceptance Criteria**:
- The architecture described in https://github.com/DRAGNLabs/MIRROR-Pipeline/issues/268#issuecomment-4338244523 is implemented
- The first implemented metric is Bits-per-byte (see #269)

morrisoncheney:
Tested with the following config file by running

python src/main.py eval --config <config_name>

```model:
  class_path: MirrorLlamaModel
  init_args:
    initialization:
      vocab_size: 8000
      hidden_size: 128
      intermediate_size: 340
      num_hidden_layers: 6
      num_attention_heads: 4
      num_key_value_heads: null

evaluator:
  class_path: MirrorEvaluator

metrics:
  bits_per_byte:
      class_path: BitsPerByteMetric
      init_args:
        data:
          class_path: WikitextDataset
          init_args:
            split: validation
            head: 50
        preprocessor:
          class_path: BPEPreprocessor
          init_args:
            file_path: /home/morriskc/groups/grp_mirror/mirror_data/datasets/church-english.txt
            vocab_size: 8000
  grad_norm:
      class_path: GradNormMetrics
  l2_norm:
      class_path: L2NormMetrics

checkpoint_path: /home/morriskc/groups/grp_mirror/nobackup/autodelete/baby_babble/end.ckpt

slurm:
  job_type: compute
  time: "01:00:00"
  gpus_per_node: "1"
  mem_per_cpu: "16G"
  output: "slurm_logs/%j.out"
  open_mode: "append"
  signal: "SIGHUP@90"
  requeue: true

device: cuda
```

Closes #303


src/mirror/metrics/bits_per_byte_metric.py
```diff
@@ -0,0 +1,54 @@
...
+                num_bytes = len(row['text'].encode('utf-8'))
+
+                loss_nats = model.training_step(batch).loss.item()
+
+                # causal LM predicts T-1 targets from T tokens
+                total_bits += loss_nats / math.log(2) * (num_tokens - 1)
+                total_bytes += num_bytes
+
+        total_bits_global = cast(torch.Tensor, fabric.all_reduce(torch.tensor(total_bits, device=fabric.device), reduce_op="sum")).item()
+        total_bytes_global = cast(torch.Tensor, fabric.all_reduce(torch.tensor(total_bytes, device=fabric.device), reduce_op="sum")).item()
```
morrisoncheney, line 56: I added these casts to fix a pylance typing error but for some reason the pyright check doesn't like them.

JansenNye (approved):
Lgtm and worked for me, note the Pylance errors in `bits_per_byte_metric.py` though

docs/mirror onboarding/mirror onboarding.md
```diff
@@ -277,6 +277,11 @@ Vim is the default editor for commit message files (e.g. git merge).
     - Useful for preparing data separately before running a training job
     - Requires `--data` and `--preprocessor` to be specified (either in the config file or as command-line arguments)
 
+- `python src/main.py eval --config <config-file>`: Run evaluation metrics on a trained model
+    - Requires a model, an evaluator, and at least one metric to be specified in the config file
+    - Optionally accepts a `checkpoint_path` (a direct path to a `.ckpt` file or FSDP checkpoint directory) to load trained weights before evaluating
+    - See [Evaluating a model](#evaluating-a-model) for a full walkthrough
```
Plyb: dead link?

also there are some references to `ExtraMetricGetter` in here that need to be updated.

src/mirror/metrics/bits_per_byte_metric.py
```diff
@@ -0,0 +1,54 @@
...
+                row: TextRow = self.data[i]
+                tokens = preprocessor.preprocess_example(row)
+                batch = preprocessor.collate([tokens])
+
+                num_tokens = len(tokens)
+                num_bytes = len(row['text'].encode('utf-8'))
+
+                loss_nats = model.training_step(batch).loss.item()
+
+                # causal LM predicts T-1 targets from T tokens
```
Plyb: this comment is correct, but it does point out a small error: num_bytes is being calculated based on the full text, not the full text minus the last token (which is only there to calculate the loss. It won't actually affect the results all that much, so idk if it's worth worrying about. At least a comment acknowledging that should be here though, if we don't actually solve it.

src/mirror/metrics/bits_per_byte_metric.py
```diff
@@ -0,0 +1,54 @@
...
+                total_bits += loss_nats / math.log(2) * (num_tokens - 1)
+                total_bytes += num_bytes
+
+        total_bits_global = cast(torch.Tensor, fabric.all_reduce(torch.tensor(total_bits, device=fabric.device), reduce_op="sum")).item()
```
Plyb, line 55: on these warnings: I *think* this is being caused by an edge case where pyright gets mad that `torch` doesn't explicitly declare `tensor` to be part of its public interface. It can be fixed by adding a line to the pyrightconfig.json that allows private imports (can't remember the exact name off the top of my head)

src/main.py
```diff
@@ -89,6 +90,34 @@ def _run(subcommand: Subcommand):
...
             preprocess(**init)
 
+        case 'eval':
+            from mirror.mirror_evaluator import MirrorEvaluator
+
+            parser = ArgumentParser()
+            parser.add_argument("--config", action=ActionConfigFile)
+            parser.add_function_arguments(evaluation, as_positional=False, skip={"model", "evaluator"})
+            parser.add_subclass_arguments(MirrorModel, "model", required=True, instantiate=False)
+            parser.add_subclass_arguments(MirrorEvaluator, "evaluator", required=True, instantiate=True)
```
Plyb: Any particular reason we're treating Evaluator specially here?

src/main.py
```diff
@@ -89,6 +90,34 @@ def _run(subcommand: Subcommand):
...
+            parser.add_argument("--device", type=str, choices=["cpu", "cuda"], default=None)
+            cfg = parser.parse_args(resolve_config_args(sys.argv[2:]))
+
+            if hasattr(cfg, 'config'):
+                del cfg.config  # pyright: ignore
+
+            init_config(cfg.device)
+            init_cfg = cfg.clone()
```
Plyb: Why the clone? (come to think of it, why is it being cloned in `fit`?)

src/main.py
```diff
@@ -89,6 +90,34 @@ def _run(subcommand: Subcommand):
...
+            init = parser.instantiate_classes(init_cfg)
+            model = instantiate_model(init.model, fabric=None)
```
Plyb: fabric=None here means that if we have a model bigger than can fit on rank-0's GPU, we'll get a cuda OOM error. You may need to call `make_fabric` before this and pass it in.

Plyb (changes requested):

src/main.py
morrisoncheney: I was just obeying the laws of `fit` lol.

docs/mirror onboarding/mirror onboarding.md
morrisoncheney: SHould be taken care of.

src/mirror/metrics/bits_per_byte_metric.py
morrisoncheney: I added a `-1` to the num_bytes. Is that all we need to resolve that? (I also deleted this `# causal LM predicts T-1 targets from T tokens` since it seems like it's unnecessary now that a correction has been made.)

src/mirror/metrics/bits_per_byte_metric.py
morrisoncheney: I tried adding something to the config. We'll see if it corrects the error (since my Pylance doesn't show those lines as having any errors).

src/main.py
morrisoncheney: After some more thought it seemed completely unnecessary to have those clones in so I just got rid of both of them.

JansenNye (changes requested):
"No parameter named num_nodes" error in trainer.py

src/mirror/metrics/bits_per_byte_metric.py
Plyb: not quite. To do this properly, we'd need to calculate how many bytes are in the text row if the final token were removed. So we'd need to figure out what the final token is, chop that off, then run the utf-8 encoding to get the bytes. Our preprocessor unfortunately isn't really set up to do that, since we'd need a way to convert from the tokens back to text.

I'm waffling on whether it's worth dealing with or if we just remove the -1 and accept that it's a little imprecise (with a comment acknowledging that). The alternatives would be to somehow restructure our preprocessors to do de-tokenization or to get hacky and just assume there is an underlying tokenizer (which there isn't always, but maybe we can assume that for this specific kind of preprocessor? idk)

Plyb (changes requested):

src/mirror/metrics/bits_per_byte_metric.py
Plyb: yeah, looks like those went away. Now we're just getting the `num_nodes` thing, which you already made a ticket for

Plyb: For some reason Pylance be default is configured not to report the private imports thing, but pyright is

Plyb (changes requested):
Just the `- 1` thing. Also @JansenNye, can you switch your review to approve if you're good with it so that GitHub doesn't block the merge?

Plyb (approved):

Plyb (approved):

PR #329, opened 2026-05-20
Implement customizable optimization

### LINKED ISSUE #190: Customizable Optimization (Gradient Accumulation)
**Acceptance Criteria**:
- users can customize how the optimization step happens, including under what conditions
- the `on_train_batch_end` callback method is split into two callback methods
  - `on_optimization_step`: for when the optimization step actually happens
  - `on_train_batch_end`: after every batch iteration
  - existing implementations are split between the two new methods as appropriate
- A default optimization strategy exists that works as it currently does
- A gradient accumulation optimization strategy exists

**Possible Architecture**:
- There is a `OptimizationStrategy` interface that has a method that takes in the kinds of information needed for optimization (an optimizer, the batch loss, batch index, and anything else that ends up being needed)
  - There are two implementations of this, one for how we're doing it now, one for gradient accumulation
  - `fit` takes in an `OptimizationStrategy` with a default and calls the method

**(OLD) Possible architecture**:
Actually don't do this, since it requires nested callbacks.
- Optimization strategies can be a callback with a `on_train_batch_end` method
  - We'd need some way to have the default optimization strategy be overridable.
    - Easiest way would be to special case this: We do something like having the optimization strategies inherit a `OptimizationCallback(Callback)` abstract class, then ensure only one is in the callback list. Don't love this, but it would work.
    - We could change how we deal with default callbacks:
      - maybe each callback could have a `filter` method that would allow it to remove other callbacks from the callback list. This would generalize the `singleton` thing as well as this problem

JansenNye:
Tested by running a smoke test confirming that `DefaultOptimizationStrategy` steps every batch and `GradientAccumulationStrategy(N=4)` only steps every 4th batch with weights changing accordingly
Closes #190

### DIFF
--- a/src/mirror/trainer.py
+++ b/src/mirror/trainer.py
@@ -18,6 +18,8 @@ from mirror.callbacks.print_step_callback import PrintStepCallback
 from mirror.callbacks.progress_callback import ProgressCallback
 from mirror.callbacks.wandb_callback import WandbCallback
 from mirror.checkpoint_identifier import CheckpointIdentifier
+from mirror.optimization.default_optimization_strategy import DefaultOptimizationStrategy
+from mirror.optimization.optimization_strategy import OptimizationStrategy
 from mirror.schedulers.configure_scheduler import ConfigureScheduler
 from mirror.config import RuntimeEnvironment, get_config
 from mirror.datasets.mirror_dataset import MirrorDataset, preprocess_dataset
@@ -97,7 +99,9 @@ class Trainer[RawT, ProcessedT, BatchT, ModelOutputT]:
             val_check_interval: int = 1,
             configure_scheduler: ConfigureScheduler | None = None,
             shuffle: bool = True,
+            optimization_strategy: OptimizationStrategy | None = None,
     ):
+        optimization_strategy = optimization_strategy or DefaultOptimizationStrategy()
         training_run_id = datetime.datetime.now().isoformat()
         rank_zero_log(self.fabric, f"Training run ID: {training_run_id}\n")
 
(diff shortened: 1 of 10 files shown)


morrisoncheney (approved):
lgtm (lovely, goated, tubular, magical)

src/mirror/trainer.py
```diff
@@ -110,7 +114,7 @@ def fit(
 
 
         dataloader = make_dataloader(dataset, preprocessor, batch_size, do_preprocess,
-                                    shuffle, num_nodes=self.num_nodes, fabric=self.fabric)
+                                    shuffle, fabric=self.fabric)
```
Plyb, line 117: Was this just a bug from another ticket?

src/mirror/trainer.py
```diff
@@ -176,20 +182,32 @@ def fit(
...
+                    optimizer_stepped = optimization_strategy.step(
+                        fabric=self.fabric,
+                        optimizer=optimizer,
+                        loss=train_step_output.loss,
+                        batch_idx=batch_idx,
+                    )
+                    if optimizer_stepped:
+                        optimization_step += 1
+                        if scheduler is not None:
+                            scheduler.step()
```
Plyb, line 196: The scheduler is configured to schedule with a max number of times `.step()` will be called. This desyncs the number we are configuring it with and the actual number of times we call it. We should either:

1. configure the scheduler to use the number of optimizer steps (might be hard in general. I could imagine an optimization strategy that doesn't know up front how many times it will be called, but maybe that's just rare).
2. call the scheduler's `.step` every training step, even if we don't call optimizer.step. Idk if that would break things though...

also, out of scope for this ticket, but should we be resetting or reconstructing the scheduler every epoch?

Plyb (changes requested):

src/mirror/trainer.py
JansenNye: Yeah, somewhere along the way it got merged to main

src/mirror/trainer.py
JansenNye: Apparently stuff like cosine annealing schedulers are meant to span the whole run instead of being broken, so we should probably leave it alone

Plyb (approved):

PR #333, opened 2026-05-21
Implement CSVDataset

### LINKED ISSUE #254: Develop Q/A fine-tuning dataset
**Acceptance Criteria**:
- we have our own fine-tuning dataset that can fine-tune a base model (of size >100M parameters) to a Instruct-tuned (chatbot) model
  - the success of this fune-tuning dataset is verified
  - This dataset includes examples demonstrating:
    - abstention on ambiguity
    - citations (RAG usage)

**Notes**
- this does not require that we have chat templates or anything like that
- We should probably just test this on existing base models so we don't have to worry about fluency. If it works on a 1B parameter model, it should work on a 100M parameter model.
- We should be getting a Q/A dataset from Bam of Solace at some point.

JansenNye:
Tested by running a local training run and confirming that the Solace fine-tuning dataset loads properly and trains on a small Llama model 
Closes #254

### DIFF
--- a/src/mirror/datasets/csv_dataset.py
+++ b/src/mirror/datasets/csv_dataset.py
@@ -1,0 +1,91 @@
+from pathlib import Path
+from typing import cast
+
+from datasets import Dataset, load_dataset
+
+from mirror.datasets.mirror_dataset import MirrorDataset
+from mirror.types import PromptResponseRow, TextRow
+
+
+class CsvDataset(MirrorDataset[TextRow]):
+    """
+    Loads a CSV and flattens each row to a single text string via a Python
+    format-string template referencing column names. Example:
+
+        CsvDataset(file_path, template="{text}")  # single-column CSV
+        CsvDataset(file_path, template="Q: {query}\\nA: {response}")
+    """
+
+    @property
+    def ds(self) -> Dataset:
+        return self._ds
+
+    def __init__(
+        self,
+        file_path: str | Path,
+        template: str = "{text}",
+        head: int | None = None,
+    ) -> None:
+        super().__init__()
+        self._template = template
+        self._ds = cast(Dataset, load_dataset("csv", data_files=str(file_path), split="train"))
+        self._ds = self._ds.filter(lambda row: len(self._render(row)) > 0)
+        if head:
+            self._ds = self._ds.select(range(head))
+
+    def _render(self, row: dict) -> str:
+        return self._template.format(**{k: ("" if v is None else v) for k, v in row.items()})
+
+    def to_row_type(self, ds_row: dict) -> TextRow:
+        return TextRow(text=self._render(ds_row))
+
+    def __len__(self) -> int:
+        return len(self.ds)
+
+
+class CsvSftDataset(MirrorDataset[PromptResponseRow]):
+    """
+    Loads a CSV for supervised fine-tuning. Each row produces a
+    PromptResponseRow by rendering separate prompt and response templates
+    against the CSV columns. Example for a (query, response, sources) CSV:
+
+        CsvSftDataset(
+            file_path,
+            prompt_template="Query: {query}\\nSources: {sources}\\nResponse: ",
+            response_template="{response}",
+        )
+    """
+
+    @property
+    def ds(self) -> Dataset:
+        return self._ds
+
+    def __init__(
+        self,
+        file_path: str | Path,
+        prompt_template: str,
+        response_template: str,
+        head: int | None = None,
+    ) -> None:
+        super().__init__()
+        self._prompt_template = prompt_template
+        self._response_template = response_template
+        self._ds = cast(Dataset, load_dataset("csv", data_files=str(file_path), split="train"))
+        self._ds = self._ds.filter(lambda row: len(self._render_response(row)) > 0)
+        if head:
+            self._ds = self._ds.select(range(head))
+
+    def _render(self, template: str, row: dict) -> str:
+        return template.format(**{k: ("" if v is None else v) for k, v in row.items()})
+
+    def _render_response(self, row: dict) -> str:
+        return self._render(self._response_template, row)
+
+    def to_row_type(self, ds_row: dict) -> PromptResponseRow:
+        return PromptResponseRow(
+            prompt=self._render(self._prompt_template, ds_row),
+            response=self._render_response(ds_row),
+        )
+
+    def __len__(self) -> int:
+        return len(self.ds)
--- a/src/mirror/datasets/csv_dataset.py
+++ b/src/mirror/datasets/csv_dataset.py
@@ -47,11 +47,14 @@ class CsvSftDataset(MirrorDataset[PromptResponseRow]):
     """
     Loads a CSV for supervised fine-tuning. Each row produces a
     PromptResponseRow by rendering separate prompt and response templates
-    against the CSV columns. Example for a (query, response, sources) CSV:
+    against the CSV columns. The prompt is the user-message content and the
+    response is the assistant-message content; role markers and special
+    tokens are added later by the SFT preprocessor's chat template (when
+    the tokenizer has one). Example for a (query, response, sources) CSV:
 
         CsvSftDataset(
             file_path,
-            prompt_template="Query: {query}\\nSources: {sources}\\nResponse: ",
+            prompt_template="{query}\\nSources: {sources}",
             response_template="{response}",
         )
     """
@@ -71,20 +74,19 @@ class CsvSftDataset(MirrorDataset[PromptResponseRow]):
         self._prompt_template = prompt_template
         self._response_template = response_template
         self._ds = cast(Dataset, load_dataset("csv", data_files=str(file_path), split="train"))
-        self._ds = self._ds.filter(lambda row: len(self._render_response(row)) > 0)
+        self._ds = self._ds.filter(
+            lambda row: len(self._render(self._response_template, row)) > 0
+        )
         if head:
             self._ds = self._ds.select(range(head))
 
     def _render(self, template: str, row: dict) -> str:
         return template.format(**{k: ("" if v is None else v) for k, v in row.items()})
 
-    def _render_response(self, row: dict) -> str:
-        return self._render(self._response_template, row)
-
     def to_row_type(self, ds_row: dict) -> PromptResponseRow:
         return PromptResponseRow(
             prompt=self._render(self._prompt_template, ds_row),
-            response=self._render_response(ds_row),
+            response=self._render(self._response_template, ds_row),
         )
 
     def __len__(self) -> int:
--- a/src/mirror/datasets/csv_dataset.py
+++ b/src/mirror/datasets/csv_dataset.py
@@ -2,9 +2,11 @@ from pathlib import Path
 from typing import cast
 
 from datasets import Dataset, load_dataset
+from typed_datasets import TypedDataset
 
 from mirror.datasets.mirror_dataset import MirrorDataset
 from mirror.types import PromptResponseRow, TextRow
+from mirror.util import _ds_cache_path_context
 
 
 class CsvDataset(MirrorDataset[TextRow]):
@@ -17,7 +19,7 @@ class CsvDataset(MirrorDataset[TextRow]):
     """
 
     @property
-    def ds(self) -> Dataset:
+    def ds(self) -> TypedDataset[TextRow]:
         return self._ds
 
     def __init__(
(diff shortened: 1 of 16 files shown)


morrisoncheney (approved):
money money money money

src/mirror/datasets/csv_dataset.py
```diff
@@ -0,0 +1,91 @@
...
+
+    @property
+    def ds(self) -> Dataset:
+        return self._ds
+
+    def __init__(
+        self,
+        file_path: str | Path,
+        prompt_template: str,
+        response_template: str,
```
Plyb, line 73: most tokenizers for hf causal lms have a "chat template" that distinguishes between the prompt and response. Instead of reinventing the wheel, it may be worth using hf's established formats for chat conversations.

https://huggingface.co/docs/transformers/en/chat_templating

The idea would be to return rows in hf's dictionary format (role/content) and let the preprocessor do the work of formatting it into a conversation. That said, the sources thing makes our jobs a bit harder. They should be in the prompt as your docstring points out. I guess the thing to do here would be to keep the template strings like you have them, but remove the "Query:" and "Response:" labels, since those will be handled by the chat template?

src/mirror/datasets/csv_dataset.py
```diff
@@ -0,0 +1,91 @@
...
+    def _render(self, template: str, row: dict) -> str:
+        return template.format(**{k: ("" if v is None else v) for k, v in row.items()})
+
+    def _render_response(self, row: dict) -> str:
+        return self._render(self._response_template, row)
+
+    def to_row_type(self, ds_row: dict) -> PromptResponseRow:
+        return PromptResponseRow(
+            prompt=self._render(self._prompt_template, ds_row),
+            response=self._render_response(ds_row),
```
Plyb: optional nit: mildly confusing that response gets a dedicated helper function but prompt doesn't

src/mirror/models/mirror_gpt_model.py
```diff
@@ -8,38 +7,38 @@
...
+from mirror.preprocessors.mirror_preprocessor import MirrorPreprocessor
+from mirror.types import AttentionMaskBatch, LabeledTokens, LabelsBatch, Loss, TextRow, TokenBatch, TrainStepOutput
 
 
 hf_model_name = "openai-community/gpt2"
 
 class MirrorGPTModel(
-    MirrorModel[TextRow, TokenTensor, tuple[TokenBatch, AttentionMaskBatch], None],
-    HFWhiteboxTransformer    
+    MirrorModel[TextRow, LabeledTokens, tuple[TokenBatch, AttentionMaskBatch, LabelsBatch], None],
```
Plyb: optional: might be worth giving the `tuple[TokenBatch, AttentionMaskBatch, LabelsBatch]` thing its own name. Maybe `StandardBatch` or `LLMBatch`.

src/mirror/models/mirror_gpt_model.py
```diff
@@ -8,38 +7,38 @@
...
+    HFWhiteboxTransformer
 ):
-    def __init__(self, weights: Literal["pretrained", "random"] = "pretrained") -> None:
+    def __init__(
+        self,
+        weights: Literal["pretrained", "random"] = "pretrained",
+        preprocessor: MirrorPreprocessor | None = None,
```
Plyb: Why is adding the `preprocessor` here required? Does our `preprocessor` override system not work for this use case?

src/mirror/models/model_util.py
```diff
@@ -7,9 +7,10 @@
 
 from mirror.download_util import assert_can_download
 from mirror.models.mirror_model import MirrorModel
+from mirror.types import IGNORE_ID
 from mirror.util import mirror_data_path
 
-IGNORE_ID = -100 
+__all__ = ["IGNORE_ID", "load_hf_model", "load_hf_config", "build_causal_lm", "instantiate_model"]
```
Plyb: why is this needed?

src/mirror/preprocessors/mirror_sft_preprocessor.py
```diff
@@ -0,0 +1,44 @@
...
+from mirror.types import AttentionMaskBatch, IGNORE_ID, LabeledTokens, LabelsBatch, PromptResponseRow, TokenBatch
+
+
+class MirrorSftPreprocessor(
+    MirrorPreprocessor[PromptResponseRow, LabeledTokens, tuple[TokenBatch, AttentionMaskBatch, LabelsBatch]]
+):
+    """
+    Tokenizes prompt and response separately so the loss can be masked on
+    prompt tokens. Emits one input_ids sequence (prompt + response) and a
+    matching labels sequence with IGNORE_ID over prompt positions.
```
Plyb: yeah, I'd recommend relying on the build in chat templates and `return_assistant_tokens_mask=True` for this.

src/mirror/preprocessors/preprocessor_util.py
```diff
@@ -37,10 +38,24 @@ def load_hf_tokenizer(
...
+
+    padded_len = input_ids.shape[1]
+    pad_left = getattr(tokenizer, "padding_side", "right") == "left"
+    labels = torch.full((len(examples), padded_len), IGNORE_ID, dtype=torch.long, device=device)
+    for i, e in enumerate(examples):
+        row = torch.as_tensor(e["labels"], dtype=torch.long, device=device)
+        if pad_left:
+            labels[i, padded_len - row.shape[0]:] = row
+        else:
+            labels[i, :row.shape[0]] = row
```
Plyb: Out of curiosity, is there not a (clean) way for us to get `tokenizer.pad` to do this for us? Seems like it's basically duplicating its logic.

Plyb (changes requested):

src/mirror/datasets/csv_dataset.py
JansenNye: Yeah I'll nuke the response helper

src/mirror/models/mirror_gpt_model.py
JansenNye: Oh, you're right, it does

src/mirror/models/model_util.py
JansenNye: Oh yeah that's dead code

src/mirror/preprocessors/preprocessor_util.py
JansenNye: Yeah, let me do that

src/mirror/preprocessors/mirror_instruct_preprocessor.py
```diff
@@ -1,17 +1,22 @@
...
 class MirrorSftPreprocessor(
-    MirrorPreprocessor[PromptResponseRow, LabeledTokens, tuple[TokenBatch, AttentionMaskBatch, LabelsBatch]]
+    MirrorPreprocessor[PromptResponseRow, LabeledTokens, StandardBatch]
 ):
     """
     Tokenizes prompt and response separately so the loss can be masked on
-    prompt tokens. Emits one input_ids sequence (prompt + response) and a
-    matching labels sequence with IGNORE_ID over prompt positions.
+    prompt tokens. If the tokenizer has a chat template (most HF instruct
+    models do), wraps prompt/response as a user/assistant conversation so
```
Plyb, line 16: Not for this ticket, but this does remind me: we'll need to figure out how to do this for our BPEPreprocessor. Probably shouldn't be too hard. I'll make a ticket for it.

src/mirror/preprocessors/mirror_sft_preprocessor.py
```diff
@@ -36,7 +36,32 @@ def preprocess_example(self, example: PromptResponseRow) -> LabeledTokens:
...
+        response_ids = self._tokenizer.encode(example['response'], add_special_tokens=False)
+        eos_id = self._tokenizer.eos_token_id
+        if eos_id is not None and (len(response_ids) == 0 or response_ids[-1] != eos_id):
+            response_ids = [*response_ids, eos_id]
+        return cast(TokenTensor, list(prompt_ids)), cast(TokenTensor, list(response_ids))
+
+    def _chat_token_ids(self, messages: list[dict], add_generation_prompt: bool = False) -> list[int]:
+        """
+        apply_chat_template returns either a list[int] or a BatchEncoding
+        (depending on tokenizer backend / transformers version). Normalize
```
Plyb: re "transformers version": Our transformers version is fixed. Is this saying that if we update it, the return type will change? Or that versions older than ours have a different return type? If the latter, we don't need to worry about that. If the former, I think we only worry about that if we know there is a future version that does it differently than our version.

src/mirror/preprocessors/mirror_instruct_preprocessor.py
```diff
@@ -36,7 +36,32 @@ def preprocess_example(self, example: PromptResponseRow) -> LabeledTokens:
...
+            prompt_ids = self._chat_token_ids(prompt_msgs, add_generation_prompt=True)
+            full_ids = self._chat_token_ids(full_msgs)
+            return prompt_ids, full_ids[len(prompt_ids):]
+
+        prompt_ids = self._tokenizer.encode(example['prompt'], add_special_tokens=True)
+        response_ids = self._tokenizer.encode(example['response'], add_special_tokens=False)
+        eos_id = self._tokenizer.eos_token_id
+        if eos_id is not None and (len(response_ids) == 0 or response_ids[-1] != eos_id):
+            response_ids = [*response_ids, eos_id]
+        return cast(TokenTensor, list(prompt_ids)), cast(TokenTensor, list(response_ids))
```
Plyb, line 60: `apply_chat_template` has a kwarg `return_assistant_tokens_mask`. I believe that basically would work as a replacement for what you're doing here and above with multiplying the IGNORE_ID

Plyb (changes requested):

src/mirror/preprocessors/mirror_instruct_preprocessor.py
```diff
@@ -38,11 +38,19 @@ def preprocess_example(self, example: PromptResponseRow) -> LabeledTokens:
...
+            ]
+            out = cast(BatchEncoding, self._tokenizer.apply_chat_template(
+                full_msgs,
+                return_dict=True,
+                return_assistant_tokens_mask=True,
+            ))
+            full_ids = list(cast(list[int], out["input_ids"]))
+            mask = cast(list[int], out["assistant_tokens_mask"])
+            split = next((i for i, m in enumerate(mask) if m == 1), len(full_ids))
+            return full_ids[:split], full_ids[split:]
```
Plyb, line 53: This brings up the point that we're only supporting single-message "conversations" here. Probably not a problem at the moment, considering that's all the data we have, but good to keep in mind.

Plyb (approved):
approved, but merge conflicts

src/mirror/datasets/csv_dataset.py
```diff
@@ -0,0 +1,93 @@
...
+    """
+    Loads a CSV and flattens each row to a single text string via a Python
+    format-string template referencing column names. Example:
+
+        CsvDataset(file_path, template="{text}")  # single-column CSV
+        CsvDataset(file_path, template="Q: {query}\\nA: {response}")
+    """
+
+    @property
+    def ds(self) -> Dataset:
```
Plyb: This pyright warning (and the others) needs to be fixed. We're using `TypeDataset`s now. Also note that (since we're in a transition phase), the transformation of the rows to `TextRow`s should be done in the constructor, not in `to_row_type`. `to_row_type` should just be a no-op

Plyb (changes requested):

src/mirror/preprocessors/mirror_sft_preprocessor.py
```diff
@@ -0,0 +1,67 @@
+from typing import cast
+
+from transformers import BatchEncoding, PreTrainedTokenizerBase
+
+from mirror.preprocessors.mirror_preprocessor import MirrorPreprocessor
+from mirror.preprocessors.preprocessor_util import collate_tokens, load_hf_tokenizer
+from mirror.types import IGNORE_ID, LabeledTokens, PromptResponseRow, StandardBatch, TokenTensor
+
+
+class MirrorSftPreprocessor(
```
Plyb: also optional nit: I'm not sure I'd call this `Sft` preprocessor. SFT is a wider concept than what this is doing, which is focused on what is normally called "Instruct" tuning. So maybe `Instruct` or `Chat`.

src/mirror/datasets/csv_dataset.py
```diff
@@ -27,40 +29,41 @@ def __init__(
         head: int | None = None,
     ) -> None:
         super().__init__()
-        self._template = template
-        self._ds = cast(Dataset, load_dataset("csv", data_files=str(file_path), split="train"))
-        self._ds = self._ds.filter(lambda row: len(self._render(row)) > 0)
-        if head:
-            self._ds = self._ds.select(range(head))
+        tmpl = template
```
Plyb, line 32: optional nit: any particular reason for this? It's only used in the one place, so this doesn't even save on characters in total.

Plyb (approved):

PR #338, opened 2026-05-22
Implement Mirror Inference Model

### LINKED ISSUE #196: mirror inference model
This is a slight change in architecture from how model outputs currently work. We want models to be able to have outputs besides loss. This is currently being done via the `TrainStepOutput` wrapper object returned by `training_step`. However, It doesn't make a lot of sense to always have the model output a loss if you're doing inference. Furthermore, it's weird to call "training_step" if you're wanting to do inference.

Let's create a new interface (an abstract class without any implementations) that defines a `forward` method, with a signature something like `def forward(self, batch: BatchT) -> ModelOutputT`. We'll have our existing models implement this interface (reducing code duplication as much as possible, but I think both methods will basically just use the `WhiteboxTransformerExecutor`). Note that this new interface is *not* a subclass of `MirrorModel`, it should be separate. The existing models should inherit from both this new interface and `MirrorModel`. At that point, we can have `training_step` just return `Loss` again and get rid of `TrainStepOutput`, since anywhere that wants to do inference can just use this new interface.
- should we rename `MirrorModel` to `TrainableModel` or something? Maybe not in this ticket (since that would be a pretty big change), but let me know your thoughts.

This new interface is intended for usage with inference (`InferenceModel`?), so it will also need a preprocessor, like `MirrorModel`. The cleanest way to do this imo is to factor out the `preprocessor` property into its own `HasPreprocessor` interface, which both `MirrorModel` and `InferenceModel` extend.

**Acceptance Criteria**
- A new interface exists that allows inference
  - The existing models implement this interface
- `training_step` only returns `Loss` again.

JansenNye:
```python
import torch
from mirror.interventions.intervention import Intervention
from mirror.models.configuration_llama import LlamaConfig
from mirror.models.inference_model import InferenceModel
from mirror.models.mirror_llama_model import MirrorLlamaModel
from mirror.models.placeholder_model import PlaceholderModel
from mirror.models.trainable_model import TrainableModel

def assert_scalar_loss(loss: object) -> None:
    assert isinstance(loss, torch.Tensor), f"expected Tensor, got {type(loss).__name__}"
    assert loss.dim() == 0 or loss.numel() == 1, f"expected scalar loss, got shape {loss.shape}"

def make_token_batch(vocab_size: int) -> tuple[torch.Tensor, torch.Tensor]:
    input_ids = torch.randint(0, vocab_size, (2, 8), dtype=torch.long)
    attention_mask = torch.ones_like(input_ids)
    return input_ids, attention_mask

def test_placeholder() -> None:
    model = PlaceholderModel()
    assert isinstance(model, TrainableModel)
    assert isinstance(model, InferenceModel)

    batch = make_token_batch(vocab_size=10)
    assert_scalar_loss(model.training_step(batch))
    assert model.forward(batch) is None
    print("PlaceholderModel: OK")

def test_llama_tiny() -> None:
    config = LlamaConfig(vocab_size=64, hidden_size=32, intermediate_size=64,
                        num_hidden_layers=2, num_attention_heads=2, num_key_value_heads=2)
    model = MirrorLlamaModel(initialization=config, seed=0)
    assert isinstance(model, TrainableModel)
    assert isinstance(model, InferenceModel)

    batch = make_token_batch(vocab_size=64)
    logits = model.forward(batch)
    assert isinstance(logits, torch.Tensor), f"forward returned {type(logits).__name__}"
    assert logits.shape == (2, 8, 64), f"unexpected logits shape {logits.shape}"

    loss = model.training_step(batch)
    assert_scalar_loss(loss)
    loss.backward()
    assert any(p.grad is not None for p in model.parameters()), "no parameter received gradients"
    print("MirrorLlamaModel (tiny random init): OK")

def test_intervention_proxies_training_step() -> None:
    model = PlaceholderModel()
    wrapped = Intervention(model)
    assert isinstance(wrapped, TrainableModel)

    batch = make_token_batch(vocab_size=10)
    assert_scalar_loss(wrapped.training_step(batch))
    assert wrapped.preprocessor is model.preprocessor
    print("Intervention(PlaceholderModel): OK")

def test_jsonargparse_resolves_trainable_model() -> None:
    from jsonargparse import ArgumentParser, Namespace
    from mirror.models.model_util import instantiate_model
    parser = ArgumentParser()
    parser.add_subclass_arguments(TrainableModel, "model", required=True, instantiate=False)
    ns = Namespace(model=Namespace(
            class_path="mirror.models.placeholder_model.PlaceholderModel",
            init_args=Namespace())
    )
    model = instantiate_model(ns.model, fabric=None)
    assert isinstance(model, PlaceholderModel)
    print("jsonargparse + instantiate_model: OK")

if __name__ == "__main__":
    test_placeholder()
    test_intervention_proxies_training_step()
    test_jsonargparse_resolves_trainable_model()
    test_llama_tiny()
    print("\nAll smoke tests passed.")
```

Closes #196


morrisoncheney (approved):
lgtm (like great typed-code must-look)

src/mirror/callbacks/wandb_callback.py
```diff
@@ -66,7 +66,7 @@ def on_train_batch_end(
         self,
         *,
         fabric: Fabric,
-        model: MirrorModel,
+        model: TrainableModel,
```
Plyb, line 71: Not for this ticket, but let's create a new one to add the type arguments onto TrainableModel here

src/mirror/models/trainable_model.py
```diff
@@ -0,0 +1,23 @@
...
+    @abstractmethod
+    def training_step(self, batch: BatchT) -> Loss:
+        pass
+
+    @abstractmethod
+    def configure_optimizers(self) -> Optimizer:
+        pass
+
+    def mlp_modules(self) -> list[nn.Module]:
+        return []
```
Plyb, line 24: We should think more about where these intervention-usage methods should go. Should they be in their own interface? Should that be a Protocol maybe? Gotta think more about this. But fine as it is right now.

Plyb (changes requested):
Just merge conflicts and a couple of thoughts for future tickets

Plyb (changes requested):
warnings are showing some additional usages of MirrorModel that need to be changed

JansenNye:
Note: I had Claude help resolve these merge conflicts and it noted that it made the judgment call of making `MirrorMetric` non-generic, as opposed to the the alternative of keeping main's `[RawT: Mapping[str, Any], ProcessedT, BatchT]` parametrization. That would give us more type precision for bits_per_byte, but more boilerplate on grad/l2 which don't use the params. If we want the generic version we can switch that back

Plyb:
> Note: I had Claude help resolve these merge conflicts and it noted that it made the judgment call of making `MirrorMetric` non-generic, as opposed to the the alternative of keeping main's `[RawT: Mapping[str, Any], ProcessedT, BatchT]` parametrization. That would give us more type precision for bits_per_byte, but more boilerplate on grad/l2 which don't use the params. If we want the generic version we can switch that back

It's probably fine? Does BitsPerByte even use the type parameters really?

Plyb (approved):
approved, but merge conflicts

PR #341, opened 2026-05-29
Initial implementation of Inference

### LINKED ISSUE #331: inference
Acceptance criteria:
-  there is an `infer` subcommand that mirrors the functionality of our branch 86 infer subcommand, taking in a model, preprocessor, and some text and then running inference to see what the model would predict the next text after that to be.
  - This may be difficult to get to work with any generic model, and some changes may need to be made to the `MirrorModel` structure, so there may be some R&D involved in this ticket to see how feasible it is to allow for inference with any type of model.

morrisoncheney:
This is my first attempt at making inference work for the pipeline. I created InferenceFriendly classes to be added to models that we want to make work for inference. The InferenceFriendlyModel class has nn.Module to get rid of a bunch of Pylance errors.

Tested by running the following config:

`python src/main.py infer --config infer_test.yaml`

```
model:
  class_path: MirrorGPTModel
  init_args:
    weights: pretrained

text: "Back in my day I used to"
num_tokens: 50

temperature: 0.85
top_p: 0.95
# top_k: 40
repetition_penalty: 1.1

slurm:
  job_type: compute
  time: "01:00:00"
  nodes: 1
  gpus_per_node: "1"
  # mem_per_cpu: "128G"
  output: "slurm_logs/%j.out"
  open_mode: "append"
  signal: "SIGHUP@90"
  requeue: true

device: cuda
```


src/mirror/models/inference_model.py
```diff
@@ -0,0 +1,17 @@
...
+from transformers import PreTrainedModel
+
+from mirror.models.whitebox_transformers.hf_whitebox_transformers import HFWhiteboxTransformer
+from mirror.preprocessors.mirror_preprocessor import InferenceFriendlyPreprocessor
+
+
+class InferenceFriendlyModel(nn.Module, HFWhiteboxTransformer, ABC):
+    @property
+    @abstractmethod
+    def hf_model(self) -> PreTrainedModel: ...
```
Plyb: `HFWhiteboxTransformer` already declares this property. Redefining it here doesn't break anything, but it is redundant. Limiting redundant code allows us to always assume code that we are reading is serving some purpose. If we can't assume that, it becomes harder to understand the code because you never know what is important and what isn't. So, let's remove this from the interface.

src/mirror/models/mirror_model.py
```diff
@@ -5,7 +5,7 @@
 from mirror.preprocessors.mirror_preprocessor import MirrorPreprocessor
 
 
-class MirrorModel[RawT, ProcessedT, BatchT, ModelOutputT](ABC, nn.Module):
+class MirrorModel[RawT, ProcessedT, BatchT, ModelOutputT](nn.Module, ABC):
```
Plyb: just curious, any reason for this change?

src/mirror/predictor.py
```diff
@@ -0,0 +1,57 @@
...
+from mirror.models.inference_model import InferenceFriendlyModel
+from mirror.preprocessors.mirror_preprocessor import InferenceFriendlyPreprocessor
+from mirror.util import get_device
+
+class Predictor:
+    def predict(
+            self,
+            model: InferenceFriendlyModel,
+            text: str,
+            num_tokens: int,
```
Plyb: optional
```suggestion
            max_new_tokens: int,
```

matches the hf parameter name, so someone familiar with that will be able to understand easier. Also if we ever have end of sentence tokens on our models, this will be more accurate (since hf's pipeline might generate less than the max)

src/mirror/predictor.py
```diff
@@ -0,0 +1,57 @@
...
+        device = get_device()
+
+        if checkpoint_path is not None:
+            model_state = model.state_dict()
+            dcp.load(  # type: ignore[attr-defined]
+                state_dict={'model': model_state},
+                checkpoint_id=str(checkpoint_path),
+                no_dist=True,
+            )
+            model.load_state_dict(model_state)
```
Plyb: This makes me a little nervous that we're using a different model loading strategy here than we're using other places (with the fabric-style loading). I understand we're not loading everything, like the optimizer and such, which is probably why you went with this. Can you give me any assurances that the two will always behave the same for our purposes here though?

By using two different loading systems, we increase our "bug surface area". Just more places to check if something goes wrong.

src/mirror/predictor.py
```diff
@@ -0,0 +1,57 @@
...
+        model.to(device)
+        model.eval()
+
+        tokenizer = preprocessor.tokenizer if preprocessor is not None else model.preprocessor.tokenizer
+
+        pipe = hf_pipeline(
+            'text-generation',
+            model=model.hf_model,
+            tokenizer=cast(PreTrainedTokenizerFast, tokenizer),
```
Plyb: Does the `hf_pipeline` require a `PreTrainedTokenizerFast`? If that's the case, let's have the `InferenceFriendlyPreprocessor` require a fast tokenizer instead of just assuming that the tokenizer we are being handed is a fast one.

Problem to avoid: currently someone could define an InferenceFriendlyPreprocessor with a non-fast tokenizer and the type checker wouldn't warn them, but this would fail (if it is true that `hf_pipeline` requires a fast one)

src/mirror/predictor.py
```diff
@@ -0,0 +1,57 @@
...
+            device=device,
+        )
+
+        do_sample = temperature != 1.0 or top_p is not None or top_k is not None
```
Plyb, line 43: quick check on this: Is temperature=1.0 actually deterministic? I thought that temp=0.0 was the deterministic threshold, but I could be wrong.

src/mirror/predictor.py
```diff
@@ -0,0 +1,57 @@
...
+        generation_kwargs: dict = dict(
```
Plyb, line 44: does hf have a built in type that is more specific than just `dict` here? If not, it's probably not worth worrying about, but if so, it's easy let's just use that.

Advantage: more specific types help future developers know what assumptions they have to satisfy, and what assumptions they can rely on. Types are a kind of assumption, so more specific is better.

src/mirror/subcommands.py
```diff
@@ -48,3 +49,30 @@ def preprocess(
         preprocessor: MirrorPreprocessor,
 ) -> None:
     preprocess_dataset(data, preprocessor.preprocess_example)
+
+
+def infer(
+        model: InferenceFriendlyModel,
+        text: str,
+        num_tokens: int,
```
Plyb: optional:
```suggestion
        max_new_tokens: int,
```

similar here

src/mirror/subcommands.py
```diff
@@ -48,3 +49,30 @@ def preprocess(
...
+        checkpoint_path: str | None = None,
+        preprocessor: InferenceFriendlyPreprocessor | None = None,
+        temperature: float = 1.0,
+        top_p: float | None = None,
+        top_k: int | None = None,
+        repetition_penalty: float = 1.0,
+        slurm: SlurmConfig = SlurmConfig(),
```
Plyb, line 82: Doesn't look like this `slurm` actually gets used? but it probably *should* be used? Now I don't remember where we actually do slurm launching

src/main.py
```diff
@@ -90,6 +90,29 @@ def _run(subcommand: Subcommand):
...
+            parser.add_argument("--device", type=str, choices=["cpu", "cuda"], default=None)
+            cfg = parser.parse_args(resolve_config_args(sys.argv[2:]))
+
+            if hasattr(cfg, 'config'):
+                del cfg.config  # pyright: ignore
+
+            init_config(cfg.device)
+            init_cfg = cfg.clone()
+            init = parser.instantiate_classes(init_cfg)
+            model = instantiate_model(init.model, fabric=None)
```
Plyb: optional: fabric=None might cause OOMs with larger models, but might not be worth worrying about, since I kind of doubt we're going to be using particularly large models on this pipeline.

Plyb (changes requested):

src/mirror/models/mirror_model.py
morrisoncheney: I don't remember the specifics but it was causing a really weird error. Something to do with post-compile type checking.

src/mirror/predictor.py
morrisoncheney: I believe you are correct. Should this be 0.0 then?

src/mirror/subcommands.py
morrisoncheney: It is doing the slurm submission to the supercomputer. Is that what you are asking about?

src/mirror/predictor.py
morrisoncheney: I believe we can create a GenerationConfig object, which we would have to construct and then convert to a dict. We could also create a TypedDict that has everything we want. Should we do that or try the HF GenerationConfig object?

src/mirror/predictor.py
```diff
@@ -0,0 +1,56 @@
...
+            repetition_penalty: float = 1.0,
+    ) -> str:
+        if checkpoint_path is not None:
+            if os.path.isdir(checkpoint_path):
+                import torch.distributed.checkpoint as dcp
+                model_state = model.state_dict()
+                dcp.load(state_dict={'model': model_state}, checkpoint_id=str(checkpoint_path), no_dist=True)  # type: ignore[attr-defined]
+                model.load_state_dict(model_state)
+            else:
+                fabric.load(checkpoint_path, {'model': model})
```
Plyb, line 30: related to https://github.com/DRAGNLabs/MIRROR-Pipeline/pull/341#discussion_r3335525200

Do we need the if-branch? If we have the fabric instance, shouldn't we just use it?

Reasoning: simplicity, and the fact that the branch here implies that we need it for some reason (which could be misleading to readers if we don't)

src/mirror/predictor.py
```diff
@@ -0,0 +1,56 @@
...
+
+        model.eval()
+
+        active_formatter = formatter or cast(InferFriendlyFormatter, model.formatter)
```
Plyb, line 34: Before, `InferenceModel` specifically overrided `formatter` so that it was an `InferFriendlyFormatter`, which would make this `cast` unnecessary (and would give us better type safety: you couldn't pass the type checker without including a proper `InferFriendlyFormatter`, whereas right now you could include make it fail silently). Can we re-include that override? 

https://github.com/DRAGNLabs/MIRROR-Pipeline/pull/341/changes/BASE..59fb099c447daedde803478432586a8d47c723ef#diff-bc355c1f83de35ef868cf0ca71ec8983eb5cc963222b72660f3ffb8828c6d898

src/mirror/predictor.py
Plyb: bump

src/mirror/predictor.py
Plyb: bump, but optional

src/mirror/formatters/infer_friendly_formatter.py
```diff
@@ -1,10 +1,10 @@
...
 
-from transformers import PreTrainedTokenizerBase
+from transformers import PreTrainedTokenizer, PreTrainedTokenizerFast
 
 
 class InferFriendlyFormatter(ABC):
     @property
     @abstractmethod
-    def tokenizer(self) -> PreTrainedTokenizerBase:
+    def tokenizer(self) -> PreTrainedTokenizer | PreTrainedTokenizerFast:
```
Plyb, line 9: Just out of curiosity, why was this needed?

Also, did we decide that plain `PreTrainedTokenizer` was okay? We don't need the "fast" ones?

src/mirror/predictor.py
```diff
@@ -0,0 +1,56 @@
...
+
+        pipe = hf_pipeline(
+            'text-generation',
+            model=model.hf_model,
```
Plyb, line 38: huggingface is just a mess. I just went down a rabbit hole to see if we could possibly avoid exposing the raw hf model here. Short answer: not without a heck ton of work.

Nothing to change, just complaining.

src/mirror/subcommands.py
Plyb: bump

Plyb (approved):
None of these changes are hard-blockers for merging. @morrisoncheney Could I have you create a follow up ticket with these changes, then send merge this?

src/mirror/formatters/infer_friendly_formatter.py
morrisoncheney: This is just because the pipeline accepts both kinds of tokenizer, so both types are "Infer Friendly".

morrisoncheney (commented):
Comments.

src/mirror/predictor.py
Plyb: Looking into it, `GenerationConfig` just takes a `**kwargs` as its `__init__` parameters, so we're not really buying any type safety by using it. I say skip, not worth it.

src/mirror/subcommands.py
Plyb: That is what `SlurmConfig` is for, but the instance here isn't actually being passed to anything, so I was trying to remember where slurm launching actually happens, and if we need this parameter on the function.

PR #342, opened 2026-06-01
255 initial plan for 1b model

### LINKED ISSUE #255: Initial plan for 1B model
**Acceptance Criteria**:
- We have a *rough* plan for how we are going to train a 1B parameter model, based on what we've learned from trying to train 100M parameter models.
  - This could include things like what hardware/how much data will be necessary and how we might get those resources.

morrisoncheney:
Just added 1B I think.


morrisoncheney:
I also removed num_nodes from the make_dataloader calls. num_nodes just isn't in the function signature.

Plyb (approved):

PR #344, opened 2026-06-01
252 first ssp 100m training runs kicked off

### LINKED ISSUE #252: First SSP 100M Training Runs Kicked Off
**Acceptance Criteria**:
- A training run for a 100M parameter model has been kicked off, and seems to be running successfully for the following datasets:
  - The SSP Solace data
  - The SSP BYU data
  - The SSP Internet data
  - A `MixedDataset` combining the above 3
- Each training run is specified with a very large number of epochs so that we can train to validation-loss convergence
  - if reasonable, we should also train to weight-norm convergence\\

Misc thoughts on efficient training:
- Disable cross-document attention (see last paragraph of section 2 http://arxiv.org/abs/2501.01956)
- Possible source of hyperparams: Datacomp-lm: In search of the next generation of training sets for language models. arXiv preprint arXiv:2406.11794, 2024.

JansenNye:
Tested by getting the actual training runs to work!
Closes #252


src/mirror/models/mirror_llama_model.py
```diff
@@ -51,7 +59,12 @@ def training_step(self, batch: tuple[TokenBatch, AttentionMaskBatch]) -> TrainSt
         return TrainStepOutput(loss=output.loss, output=None)
 
     def configure_optimizers(self):
-        return optim.AdamW(self.parameters())
+        return optim.AdamW(
+            self.parameters(),
+            lr=self._lr,
+            weight_decay=self._weight_decay,
+            betas=self._betas,
+        )
```
morrisoncheney, line 62: Any comment on this PyRight error @JansenNye ?

src/mirror/fabric_util.py
```diff
@@ -10,13 +10,16 @@ def make_fabric(
...
+        precision: str | None = None,
 ) -> Fabric:
+    precision_kwargs = {"precision": precision} if precision else {}
     return Fabric(
         strategy=strategy,
         devices=devices,
         num_nodes=num_nodes,
         callbacks=callbacks,
         accelerator=accelerator,
+        **precision_kwargs,
```
morrisoncheney: More joyous pyright errors!! Any idea why these are happening?

src/mirror/datasets/fineweb_dataset.py
```diff
@@ -31,6 +33,8 @@ def __init__(
             self._process,
         ))[split]
 
+        self._ds = slice_by_fraction(self._ds, start_fraction, end_fraction)
```
morrisoncheney: It would probably be good to add some sort of protection for passing in `start_fraction`, `end_fraction`, `skip`, and `head`. I'd imagine things could get problematic if the user passed in all of those.

morrisoncheney (changes requested):
Merge conflicts and a few comments.

src/mirror/models/mirror_llama_model.py
JansenNye: No idea, I'll iron it out

src/mirror/datasets/fineweb_dataset.py
JansenNye: Good idea, I'll add a separate helper that can handle any combination of those four

morrisoncheney (approved):
Looks good. One note: We are trusting users to remember to not pass in fractions and head, skip values. That seems like something to be aware of. Maybe worth adding something to the documentation about that but that might be worth making a new ticket for.

src/mirror/datasets/mixed_dataset.py
```diff
@@ -1,30 +1,46 @@
...
 from typed_datasets import TypedDataset, concatenate
 
 from mirror.datasets.mirror_dataset import MirrorDataset
 
 
 class MixedDataset[RawT: Mapping[str, Any]](MirrorDataset[RawT]):
+    # NOTE: parallel `datasets` + `weights` lists rather than the upstream
+    # `weighted_datasets: Sequence[tuple[MirrorDataset, float]]` — jsonargparse can't
+    # resolve a generic `MirrorDataset[RawT]` nested inside a tuple, so the tuple form
+    # fails to parse from YAML. See the SSP mixed config.
```
Plyb: Two comments:
1. "upstream" in my experience usually refers to "main" or the branch you're going to merge into. This comment would therefore be a bit strange to merge since it would be talking about the "upstream" sequence of tuples that no longer exists "upstream".
2. Does it work to just remove the `RawT`, or do you actually have to do two parallel lists? I would be a bit surprised if it being a sequence of tuples vs two separate lists was actually the problem.

Why prefer sequence of tuples? It guarantees same number of floats and datasets (and more closely associates them). In general, it is a good idea to make invalid states impossible to express. The ValueError helps, but still better to catch it at config write time instead of at run time.

src/mirror/models/mirror_llama_model.py
```diff
@@ -19,18 +19,26 @@ class MirrorLlamaModel(
...
+        initialization: Literal["3.2-1B", "3.2-1B-Instruct"] | dict | LlamaConfig = "3.2-1B-Instruct",
         seed: int | None = None,
+        lr: float = 1e-3,
+        weight_decay: float = 0.0,
+        betas: tuple[float, float] | list[float] = (0.9, 0.999),
     ) -> None:
         super().__init__()
         self._preprocessor = MirrorLlamaPreprocessor()
+        if isinstance(initialization, dict):
+            initialization = LlamaConfig(**initialization)
```
Plyb, line 31: Why is this needed?

src/mirror/fabric_util.py
```diff
@@ -10,13 +12,15 @@ def make_fabric(
...
         callbacks: list = [],
+        precision: str | None = None,
 ) -> Fabric:
     return Fabric(
         strategy=strategy,
         devices=devices,
         num_nodes=num_nodes,
         callbacks=callbacks,
         accelerator=accelerator,
+        precision=cast("_PRECISION_INPUT | None", precision),
```
Plyb: for my own sake, why does the type here need to be in quotes? Because it uses `|`?

src/mirror/fabric_util.py
```diff
@@ -10,13 +12,15 @@ def make_fabric(
         devices: int = 1,
         num_nodes: int = 1,
         callbacks: list = [],
+        precision: str | None = None,
```
Plyb: Could `_PRECISION_INPUT` be used here? (and in the caller locations?)

Plyb (changes requested):
A few small things + merge conflicts

src/mirror/datasets/mixed_dataset.py
JansenNye: 1. Good point haha
2. You're correct, implemented that

src/mirror/fabric_util.py
JansenNye: I thought so, but they aren't actually necessary in Python 3.12, took them out

src/mirror/models/mirror_llama_model.py
JansenNye: It has to do with the `rope_theta` issue that killed one of the SSP runs. If `initialization` were typed only as LlamaConfig, jsonargparse checks each YAML key against `LlamaConfig.__init__`, but that only names `vocab_size`/`rms_norm_eps` & jsonargparse can't see `**kwargs`, so it rejects keys like `rope_theta`. Adding `dict` to the type lets the YAML pass through & this branch turns it into a real LlamaConfig

src/mirror/fabric_util.py
JansenNye: Seems so but it takes some being careful with parsing and types

Plyb (approved):

PR #355, opened 2026-06-03
345 use preprocessed typeddatasets

### LINKED ISSUE #345: Use preprocessed `TypedDataset`s
Part of the new preprocessing design: #337 

**Acceptance Criteria**:
- No `MirrorDataset` directly uses a huggingface `Dataset`. Instead, we should use `TypedDataset`s in all location
  - `TypedDataset` is imported from https://github.com/Plyb/typed-datasets
- The backing `TypedDataset`s are fully processed so that they correspond to the `RowT`. In other words, instead of relying on `to_row_type` to do it, we actually map the dataset up front.
  - `to_row_type` becomes a no-op. It will be removed in a future ticket.
  - Any dataset size reductions (head, skip) are applied *before* the `map` for any `filter`s to prevent a full up front scan of the dataset

Plyb:
Closes #345 

## Caching note

The init-time `map`/`filter` are cached by HuggingFace and do **not** re-run on subsequent constructions — verified that two separate processes reuse the cached result (identical fingerprint, no recompute, no random-hash warning). `.select` (skip/take) writes no cache files, so reductions run outside the cache-path context; only `filter`/`map` run inside it. To keep caching effective, transform functions are kept free of run-varying captured state (module-level `to_text_row`, simple lambdas).

## Testing

Ran with the project env (`./.env/bin/python`):

- **Construction / typing**: `WikitextDataset(head=10)`, `ImdbDataset(head=10)`, `MixedDataset` of two → all construct; `ds` is a `TypedDataset`; rows stripped to exactly `{text}` (imdb's `label` dropped); `to_row_type` confirmed identity.
- **Caching**: same dataset built in two separate processes → second shows no `Map:`/`Filter:` recompute.
- **`_truncate`**: synthetic token-count dataset → keeps the expected cutoff rows (no 10BT download needed).
- **Txt streaming**: 1M-line file with blanks, `head=3` → returns exactly the first 3 non-empty rows near-instantly (no full scan); no-`head` path filters the whole file correctly.
- **`make_dataloader`**: both `do_preprocess=True` (`preprocess_dataset`) and `False` (`OnDemandPreprocessedDataset`) paths produce valid batches.
- Ran full training run with `fit`, both with and without do_preprocess=True

## Flags for reviewers

- **`preprocess_dataset` caching is unchanged / pre-existing.** The `do_preprocess=True` path emits HF's "couldn't be hashed, random hash used" warning because its local closure captures the dataset/preprocessor. This closure pattern is identical on `main` — out of scope for #345. (The init-time dataset processing that this ticket targets *does* cache.)
- **Type-bound scope.** Bounding `MirrorDataset[RawT]` to `Mapping[str, Any]` cascaded through the callback/trainer generics (~12 files). `RequeueMonitor`/`StateDict`/`MirrorModel`/`MirrorPreprocessor`/`MirrorMetric` were intentionally left unbounded since they don't feed `RawT` into a bounded type.

### DIFF
--- a/src/mirror/datasets/txt_dataset.py
+++ b/src/mirror/datasets/txt_dataset.py
@@ -2,7 +2,7 @@ from pathlib import Path
 from typing import cast
 
 from datasets import Dataset, load_dataset
-from typed_datasets import TypedDataset, load_typed
+from typed_datasets import TypedDataset
 
 from mirror.datasets.mirror_dataset import MirrorDataset
 from mirror.types import TextRow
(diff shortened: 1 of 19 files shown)


src/mirror/datasets/txt_dataset.py
```diff
@@ -19,17 +22,18 @@ def __init__(
         """
         Args:
             file_path: path to a .txt file where each line is one example.
-            head: how many examples to include. None includes the whole split.
+            head: how many examples to include. None includes the whole file.
```
morrisoncheney, line 25: good catch!

morrisoncheney (approved):
lgtm.

PR #357, opened 2026-06-04
346 break up and remove mirrorpreprocessor

### LINKED ISSUE #346: Format an entire dataset at once
part of the preprocessing pipeline revamp: #337 

Note that I pretty significantly changed the focus of this ticket to pare it down. Old description in the comments.

**Acceptance Criteria**:
- `MirrorPreprocessor`'s `preprocess_row` function is removed and replaced with `format_data`.
  - `def format_data(self, data_source: MirrorDataset[RawT]) -> TypedDataset[FormattedT]`
  - In other words, `format_data` handles the whole dataset at once, not just the rows.
  - a future ticket will update this to use `DataSource` instead of `MirrorDataset` 
- `preprocess_dataset` and `OnDemandPreprocessedDataset` are removed. Their usages are replaced with `format_data` directly
  - this also means no more `do_preprocess`. How much actual preprocessing to do is not just handled by the datasets themselves.

Plyb:
closes #346 

Tested by running the overfit test config locally.

Note that the github action for the overfit test failed, but it failed because it hit a rate limit downloading the dataset. Considering it worked locally, I think we're okay.

A couple implementation notes: 
- because we're using a `TypedDataset[ProcessedT]`, that forces everywhere with a `ProcessedT` to bound it with `Mapping[str, Any]`. Not great. I have thoughts on this that I'll discuss with @morrisoncheney soon.
- along those same lines, `TokenTensor` can't go into a `TypedDataset` raw because it isn't a dict, so we have to wrap it in a `TokenRow`. Might conflict with #254. If so, merge that one first.


src/mirror/preprocessors/mirror_gpt_preprocessor.py
```diff
@@ -18,20 +21,27 @@ def __init__(self, max_length: int | None = 2048) -> None:
...
+            if len(ids) < 2:
+                eos = tokenizer.eos_token_id
+                ids = [eos, eos] if len(ids) == 0 else [*ids, eos] # GPT causal LM loss shifts labels by 1, so seq_len=1 produces zero training targets
+            return {"input_ids": cast(list[int], ids)}
+
+        with _ds_cache_path_context():
+            return data_source.ds.map(tokenize, remove_columns=list(data_source.ds.columns))
+
+    def collate(self, examples: list[TokenRow]) -> tuple[TokenBatch, AttentionMaskBatch]:
+        return collate_tokens(self._tokenizer, [e["input_ids"] for e in examples])
```
Plyb: See #358

src/mirror/datasets/mirror_dataset.py
```diff
@@ -17,19 +15,3 @@ def to_row_type(self, ds_row: Mapping[str, Any]) -> RawT:
...
-    preprocessor_function: Callable[[RawT], ProcessedT],
-) -> Sequence[ProcessedT]:
-    def mappable_preprocessor_function(row: dict) -> dict:
-        return {"input_ids": preprocessor_function(dataset.to_row_type(row))}
-
-    with _ds_cache_path_context():
-        mapped = dataset.ds.unwrap().map(mappable_preprocessor_function)
-
-    print("Preprocessing complete.", file=stderr)
-    mapped.set_format(type="torch", columns=["input_ids"])
```
Plyb: It turns out `move_to_device` in `make_dataloader` was already handling for format setting for us (and in a more efficient way than this was doing). So we might actually get a small speed up from removing the `set_format` here.

However, `make_dataloader` only does `move_to_device` if you pass it a `fabric` instance (since `move_to_device` is unique to Fabric's `setup_dataloaders`). For now, this means that using `make_dataloader` is inconsistent wrt what device the data it returns is on. I think this is fine for now, because we're always passing `make_dataloader` a fabric instance, but external consumers (or future us) might run into problems. I'm making another ticket to deal with this so that this doesn't get blocked.

EDIT: we are also safe because `collate_tokens` does a `.to(device)`, but we shouldn't rely on that.

src/mirror/types.py
```diff
@@ -13,6 +13,10 @@ class TextLabelRow(TextRow):
     label: str
 
 
+class TokenRow(TypedDict):
+    input_ids: list[int]
+
+
 TokenTensor = list[int]
```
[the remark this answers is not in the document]
morrisoncheney: I am guessing you left this for the MCQA repo or something like that.

src/mirror/preprocessors/bpe_preprocessor.py
```diff
@@ -45,16 +47,22 @@ def __init__(self, file_path: Path, vocab_size: int) -> None:
...
+        def tokenize(row: TextRow) -> TokenRow:
+            encoding = raw_tokenizer.encode(row['text'], add_special_tokens=True)
+            ids = encoding.ids
+            if len(ids) < 2:
+                eos = raw_tokenizer.token_to_id("</s>")
+                ids = [eos, eos] if len(ids) == 0 else [*ids, eos]
+            return {"input_ids": cast(list[int], ids)}
+
+        with _ds_cache_path_context():
+            return data_source.ds.map(tokenize, remove_columns=list(data_source.ds.columns))
```
morrisoncheney, line 61: would be so awkward if someone created a dataset that had "input ids" as the original column. definitely not something to worry about though.

morrisoncheney (approved):
looks good

morrisoncheney (approved):
looks good

morrisoncheney (approved):
looks good

src/mirror/preprocessors/bpe_preprocessor.py
Plyb: afaik the output of `tokenize` would take precedence, so it would still be included

Plyb:
Can't fully test post-merge because of #363

morrisoncheney (approved):
looks good!

PR #377, opened 2026-06-11
rename preprocessor to formatter

### LINKED ISSUE #347: "process" -> "format"
Part of the preprocessing revamp: #337 

Note: pretty significantly changed in response to changes in #346 

**Acceptance Criteria**:
- "preprocess" and "process" are renamed to "format". This includes `MirrorPreprocessor`, its children, and anywhere else referencing those words.
  - justification: to emphasize that we're not really doing any data processing, we're just changing the format of the data so the model can read it properly.

Plyb:
Closes #347 

Tested by running the overfit config locally


morrisoncheney (approved):
looks good. Wow that is a crazy one.

PR #378, opened 2026-06-11
361 chat preprocessor can use custom bpe tokenizers

### LINKED ISSUE #361: Chat preprocessor can use custom bpe tokenizers
#254 introduced a chat preprocessor, but at the moment, it is restricted to only tokenizers in the hf hub. We need to be able to use our custom BPE tokenizers.

**Acceptance Criteria**:
- our custom BPE tokenizers can be used with the chat preprocessor
  - this probably also means figuring out how to add a chat template and any required special tokens to the BPE tokenizer.

morrisoncheney:
Primary changes:
Added InferFriendlyFormatter
Changed MirrorInstructFormatter to get passed in a formatter rather than an HF model name.
Adjusted BPEFormatter to have a chat template and such.

Tested with the following config:

```
data:
  class_path: CsvInstructDataset
  init_args:
    file_path: /home/morriskc/groups/grp_mirror/mirror_data/datasets/q_a_sources.csv
    prompt_template: "{query}\nSources: {sources}"
    response_template: "{response}"
    head: 20

formatter:
  class_path: MirrorInstructFormatter
  init_args:
    formatter:
      class_path: BPEFormatter
      init_args:
        file_path: /home/morriskc/groups/grp_mirror/mirror_data/datasets/church-english.txt
        vocab_size: 8000

model:
  class_path: MirrorLlamaModel
  init_args:
    initialization:
      vocab_size: 8000
      hidden_size: 128
      intermediate_size: 340
      num_hidden_layers: 6
      num_attention_heads: 4
      num_key_value_heads: null

slurm:
  job_type: compute
  time: "00:30:00"
  nodes: 1
  gpus_per_node: "1"
  output: "slurm_logs/%j.out"
  open_mode: "append"
  signal: "SIGHUP@90"
  requeue: true

epochs: 1
batch_size: 2
device: cuda
```


src/mirror/formatters/mirror_instruct_formatter.py
```diff
@@ -48,7 +52,7 @@ def _tokenize_split(self, example: PromptResponseRow) -> tuple[list[int], list[i
                 return_assistant_tokens_mask=True,
             ))
             full_ids = list(cast(list[int], out["input_ids"]))
-            mask = cast(list[int], out["assistant_tokens_mask"])
+            mask = cast(list[int], out["assistant_masks"])
```
Plyb, line 55: was this just a bug before? Or else what's the reason for this change?

src/mirror/formatters/infer_friendly_formatter.py
```diff
@@ -0,0 +1,10 @@
+from abc import ABC, abstractmethod
+
+from transformers import PreTrainedTokenizerBase
+
+
+class InferFriendlyFormatter(ABC):
```
Plyb, line 6: This is called "infer friendly", but I don't really see what this has to do with inference specifically. Can you clarify that for me?

Plyb (approved):
approved with just a couple of questions

src/mirror/formatters/mirror_instruct_formatter.py
morrisoncheney: This was a bug, yeah. Older versions of the transformer used the previous name and they changed to to the current name in like version 5 ish, which is what our pipeline uses.

src/mirror/formatters/infer_friendly_formatter.py
morrisoncheney: Yes. So I basically just copied this from the Inference branch, so this will be used both for inference and for Instruct stuff. I thought about naming it InferInstructFriendlyPreprocessor but decided not to because the name felt a little clunky (I'd be happy to change it to that thought since it is more clear).

Plyb: nah, that works for me, thanks!

PR #391, opened 2026-06-16
Changes to get the fine tuning done. 1. trainer - Throwing the global…

### LINKED ISSUE #321: QA Finetune SSP Models
Depends on the results of #322.
If lack of fluency, change this ticket to "identify likely
causes propose strategic interventions (domain-specific
models, synthetic data, etc.), and re-launch training runs."
In that case, unblock on #270 and #254 

**Acceptance Criteria**:
- We have Instruct-tuned versions of the 100M parameter models that we can chat ask questions to
  - For now, no need to do anything fancy with the infer subcommand, we can just input the special characters manually
  - Solace and BYU are the most important models to fine tune, since those are the blockers for #368 and #367

morrisoncheney:
Changes to get the fine tuning done:

1. trainer - Throwing the `global_step == None` error meant there coudl be no fine tuning the completely trained models. Perhaps it would be better if we just added a parameter to pass into Trainer that dodges this error and sets `global_step = 0` so training works that way we don't avoid wanted errors. 

2. Checkpoint callback - for convenience I added per-epoch functionality to the checkpoints. 

3. csv_dataset - added fractional splitting up of datasets.

Tested by actually running the fine tuning on the models.

### DIFF
--- a/src/mirror/callbacks/callback.py
+++ b/src/mirror/callbacks/callback.py
@@ -78,6 +78,21 @@ class Callback[RawT: Mapping[str, Any], FormattedT: Mapping[str, Any], BatchT, M
     ):
         pass
 
+    def on_epoch_end(
+            self,
+            *,
+            fabric: Fabric,
+            model: MirrorModel[RawT, FormattedT, BatchT, ModelOutputT],
+            optimizer: Optimizer,
+            training_run_id: str,
+            epoch: int,
+            epochs: int,
+            n_batches: int,
+            global_step: int,
+            optimization_step: int,
+    ):
+        pass
+
     def on_validation_epoch_end(
             self,
             *,
--- a/src/mirror/callbacks/checkpoint_callback.py
+++ b/src/mirror/callbacks/checkpoint_callback.py
@@ -10,9 +10,10 @@ from mirror.dict_types import StateDict
 class CheckpointCallback[RawT: Mapping[str, Any], FormattedT: Mapping[str, Any], BatchT, ModelOutputT](
        Callback[RawT, FormattedT, BatchT, ModelOutputT]
 ):
-    def __init__(self, every_n_training_steps: int | None = None) -> None:
+    def __init__(self, every_n_training_steps: int | None = None, every_n_epochs: int | None = None) -> None:
         super().__init__(is_singleton=True)
         self.every_n_training_steps = every_n_training_steps
+        self.every_n_epochs = every_n_epochs
 
     def on_fit_start(
             self,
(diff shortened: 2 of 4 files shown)


src/mirror/callbacks/callback.py
```diff
@@ -78,11 +78,26 @@ def on_optimization_step(
     ):
         pass
 
+    def on_epoch_end(
+            self,
+            *,
+            fabric: Fabric,
+            model: MirrorModel[RawT, FormattedT, BatchT, ModelOutputT],
```
Plyb, line 85: any idea what's going on with these warnings?

src/mirror/callbacks/checkpoint_callback.py
```diff
@@ -74,16 +75,38 @@ def on_optimization_step(
...
+    def on_epoch_end(
+            self,
+            *,
+            fabric: Fabric,
+            model: MirrorModel[RawT, FormattedT, BatchT, ModelOutputT],
+            optimizer: Optimizer,
+            training_run_id: str,
+            epoch: int,
+            epochs: int,
+            n_batches: int,
```
Plyb, line 87: optional: unused. Can remove from here and callback to keep the interfaces as small as possible. Not a big deal though.

src/mirror/trainer.py
```diff
@@ -133,16 +133,14 @@ def fit(
...
-            self.fabric.load(checkpoint.path, cast(dict[str, torch.nn.Module | torch.optim.Optimizer | Any], state))
+            self.fabric.load(checkpoint.path, cast(dict[str, torch.nn.Module | torch.optim.Optimizer | Any], state), strict=False)
 
             if state['global_step'] is None:
-                raise RuntimeError(
-                    "checkpoint_global_step cannot be None. "
-                    f"checkpoint '{checkpoint.checkpoint_name}' gave an invalid global step value."
-                    )
+                state['global_step'] = 0
+                state['optimization_step'] = 0
```
Plyb, line 140: This should work for this ticket, but we probably want to think this through a bit more.

As it stands, these changes would mean you could load any *finished* checkpoint for fine tuning, but if you tried to use an earlier checkpoint, it would include whatever step it was on at the time and try to continue the training from there, which might cause problems.

One solution: MirrorLlamaModel has an `init` parameter. Could we have that point to a checkpoint and use it to *just* load the weights? But then the constructor would need access to a fabric instance...

Yeah probably the best solution here is just to include something like a "full checkpoint state" or "just checkpoint weights" flag. idk. Thoughts?

Plyb (changes requested):
Thanks for getting this done. With those checkpoints in place, the other tickets can be considered unblocked, so we can take a second to consider the choices here more carefully.

Mostly just some questions.

PR #396, opened 2026-06-25
393 try instruct tuning to very easy task

### LINKED ISSUE #321: QA Finetune SSP Models

### LINKED ISSUE #393: Try Instruct Tuning to very easy task
It seems like our instruct tuned models (from #321) are unable to follow instructions in any meaningful sense. I see two hypotheses for why:
- They're just not capable of staying coherent long enough to answer
- Something went wrong in our training process

To distinguish between the two, let's use the instruct training process, but modify it to be something super easy that the tiniest language model should learn: The echo task (Just repeat whatever the user said and end the message.)
- evidence for why this should be an easy task: "induction heads" are an *extremely* common and strong component that LMs learn.

Note that this blocks #367 and #368 because if they can't follow instructions, there's no way to meaningfully define an unaligned/hallucinated response.

**Acceptance Criteria**:
- We have a fine-tuning dataset available to the team, where the response is just a copy of the query
  - I'd suggest throwing together a quick script to that processes the lutheran q_a_sources.csv into something that fits this format
- We have trained one of our LM on this dataset, and it is successfully completing the echo task

### LINKED ISSUE #397: fix tokenizers
```
Step 1.           Test fine-tuning the model with a tokenizer that has no special tokens.
                                    /                                       \
                        Doesn't work                                       Works
                                /                                              \
 Find a way to add special tokens                       Celebrate and then make sure tokens get 
      to tokenizer (see below).                                        added for future models.
```

Adding special tokens:
1. Try finding a way to just add a few layers to the model while keeping all of the other weights as they were, allowing us to stretch the vocab size of the tokenizer.
2. keep the vocab size the same and replace the last few with the new special tokens.
3. DO NOT DO THIS OPTION: Bite the bullet and redo pretraining with the new tokenizer.

Acceptance Criteria:
- The above stuff has been addressed.

morrisoncheney:
Tested by training several fine-tuned models on this.

Notes:
passing in strict=False into the thing in trainer was necessary for the checkpoints to work. I don't fully understand why.
yeah I think that's it for notes.

**UPDATED PR:**
This branch now has changes pertinent to several other branches. 

1. `every_n_epochs` for checkpoint callback.
2. Corrected BPEFormatter for our 100M parameter training runs. (The changes made for this were designed to be temporary. Ideally as soon as we are done with the 100M parameter training—I guess we might actually be done with that...—we should revert the changes in the BPEFormatter.) Closes #397.
3. Fractional splitting of the CSVDataset was implemented.
4. Echo Dataset was created to test instruct tuning. Closes #393.
5. A `just_checkpoint_weights` parameter was created for trainer so that it is now possible to begin training runs using an `end.ckpt`. (Previously we designed infrastructure that prohibited this as it didn't seem to make sense to continue a training run that was already complete, but this change demonstrates our understanding of the fact that the checkpoints are not just used to continue training runs, but also to store and deliver model weights.) Closes #321.

All of the above was tested by running the actual fine-tuning training runs.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,4 +1,6 @@
 {
   "python.defaultInterpreterPath": "${workspaceFolder}/.env/bin/python",
-  "python.terminal.activateEnvironment": false
+  "python.terminal.activateEnvironment": false,
+  "python-envs.defaultEnvManager": "ms-python.python:conda",
+  "python-envs.defaultPackageManager": "ms-python.python:conda"
 }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,6 +1,4 @@
 {
   "python.defaultInterpreterPath": "${workspaceFolder}/.env/bin/python",
-  "python.terminal.activateEnvironment": false,
-  "python-envs.defaultEnvManager": "ms-python.python:conda",
-  "python-envs.defaultPackageManager": "ms-python.python:conda"
+  "python.terminal.activateEnvironment": false
 }
--- a/src/mirror/callbacks/checkpoint_callback.py
+++ b/src/mirror/callbacks/checkpoint_callback.py
@@ -10,9 +10,14 @@ from mirror.dict_types import StateDict
 class CheckpointCallback[RawT: Mapping[str, Any], FormattedT: Mapping[str, Any], BatchT](
        Callback[RawT, FormattedT, BatchT]
 ):
-    def __init__(self, every_n_training_steps: int | None = None) -> None:
+    def __init__(
+            self,
+            every_n_training_steps: int | None = None,
+            every_n_epochs: int | None = None,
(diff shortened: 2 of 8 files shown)


JansenNye:
LGTM (lots of gooey truffles & marshmallows)

Plyb (approved):

PR #399, opened 2026-06-29
Implement Inference touch up

### LINKED ISSUE #392: Inference Touch Up
Acceptance Criteria:
1. Add abstract property to InferenceModel such that it requires a formatter -> InferenceFriendlyFormatter property
2. Sort out the below comment.
Image
3. Fix this comment: https://github.com/DRAGNLabs/MIRROR-Pipeline/pull/341/changes#r3335547668
4. Make the Predictor print out the checkpoint path so we know what checkpoint was used to generate a given piece of text. maybe print out the other config stuff for a given inference thing.

JansenNye:
Tested by running:
```
python src/main.py infer \
  --model MirrorGPTModel --model.weights pretrained \
  --text "Once upon a time" --max_new_tokens 20 \
  --device cpu --slurm.job_type local
```
And confirming that the output contains:
```
Inference checkpoint: none (base model weights)
Generation config: {'max_new_tokens': 20, 'do_sample': False, 'repetition_penalty': 1.0}
```
Closes #392

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -107,11 +107,11 @@ def _run(subcommand: Subcommand):
             init = parser.instantiate_classes(cfg)
 
             from mirror.config import get_config
-            from mirror.fabric_util import make_fabric
+            from mirror.fabric_util import cpu_safe_strategy, make_fabric
 
             config = get_config()
             fabric = make_fabric(
-                init.strategy,
+                cpu_safe_strategy(init.strategy, config['device']),
                 config['device'],
                 devices=init.slurm.ntasks_per_node or 1,
                 num_nodes=init.slurm.nodes or 1,
@@ -133,7 +133,7 @@ def _run(subcommand: Subcommand):
         case 'infer':
             from mirror.models.inference_model import InferenceModel
             from mirror.config import get_config
-            from mirror.fabric_util import make_fabric
+            from mirror.fabric_util import cpu_safe_strategy, make_fabric
 
             parser = ArgumentParser()
             parser.add_argument("--config", action=ActionConfigFile)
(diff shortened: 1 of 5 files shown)


morrisoncheney (approved):
looks good

src/main.py
```diff
@@ -107,11 +107,11 @@ def _run(subcommand: Subcommand):
             init = parser.instantiate_classes(cfg)
 
             from mirror.config import get_config
-            from mirror.fabric_util import make_fabric
+            from mirror.fabric_util import cpu_safe_strategy, make_fabric
 
             config = get_config()
             fabric = make_fabric(
-                init.strategy,
+                cpu_safe_strategy(init.strategy, config['device']),
```
Plyb, line 114: Slight pushback on this (and the one below). Using device=cpu seems like it would be pretty rare, so I'm not sure having this (undocumented) extra behavior here is worth it. I might be missing some use case though, let me know.

Plyb (changes requested):

src/main.py
JansenNye: Quick local/login node smoke tests are a plausible use case to me (and the reason I implemented this; I was testing out inference locally and ran into the "FSDP is only available on GPU" error). I added a log line so the downgrade is visible instead of hidden, but I think it's worth keeping the behavior since it's more consistent with `fit`, and because `--device cpu` is an exposed parser choice but the default strategy is `fsdp`, so without this, it just crashes with that "FSDP is only available on GPU" error.

Plyb: fair enough, seems good to me then!

Plyb (approved):

PR #400, opened 2026-06-30
Implement type-safety compatibility check

### LINKED ISSUE #94: Type safety in main
As of #7, there is some lack of type safety (type danger?) in main. Because the user can select models, datasets, and callbacks independently with one another, there is a possibility that they could select mutually incompatible options via the CLI. Right now, there are no guards against this, so the most likely outcome if the user does this is that they will get some hard-to-interpret exception. Ideally, we should either prevent them from selecting mutually incompatible options at all, or else give them a nice error message when the do.

**Acceptance Criteria**:
- One of the following is true (in order of preference):
  - The argument parser prevents selecting mutually incompatible options
    - jsonargparse's documentation *says* it supports generics (https://jsonargparse.readthedocs.io/en/stable/#generic-types), but maybe it only knows how to use the pre-python-3.12 syntax for it? Or maybe its a problem with `auto_parser`?
      - ah, I think this only works if you give it the concrete types when creating the parser, which we can't do (we want it to remain generic)
  - We run a type-compatibility check ourselves, and exit with a nice error message if the typecheck fails
    - I know we can get the types and type-arguments of values at runtime (this is called "reflection" if you need a search term)
    - However, I'm not sure how we'd check the type-compatibility. We'd need to check for subtypes on covariant type arguments, supertypes on contravariant type arguments, and type equivalence on invariant type arguments. There are ways to check the *subclass* relation, but that isn't the same as the *subtype* relation (see "structural" vs "nominal" typing. The "subtype" relation we're looking for is structural).
    - This might actually be impossible due to type erasure. I don't know if there is a way to link generic types (ensure that type `T` in one argument is the same type `T` as another, not just a different generic that happens to have the same name.
  - We perform a heuristic "sanity check" by passing a single datapoint from the dataset to the model and see if we get any errors. If there is an error, we give a nice error message.
    - ideally we'd do something similar for checking callback compatibility, but I don't know how we'd do that. Maybe something like running a very short, but full, training run, so we hit all the callbacks?

JansenNye:
Implemented the second option from the ticket description; option 1 is indeed not possible.
Tested by running a mock training run with mismatched parameters:
```
python src/main.py fit \
  --model MirrorGPTModel \
  --data CsvInstructDataset \
  --data.file_path dummy.csv \
  --data.prompt_template "{q}" \
  --data.response_template "{a}" \
  --slurm.job_type local --device cpu
```
And observing the output:
```Incompatible selection: 'CsvInstructDataset' and 'MirrorGPTModel' disagree on RawT ('PromptResponseRow' vs 'TextRow').```
Closes #94

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -24,6 +24,7 @@ def _run(subcommand: Subcommand):
     from mirror.models.model_util import instantiate_model
     from mirror.subcommands import evaluation, fit, format, infer
     from mirror.trainer_constructor import TrainerConstructor
+    from mirror.type_safety import check_compatibility
     from mirror.util import is_login_node, resolve_config_args
 
     # These are required so that their items can be found easily by jsonargparse without
@@ -55,6 +56,7 @@ def _run(subcommand: Subcommand):
             parser.add_subclass_arguments(TrainerConstructor, "trainer", required=False, instantiate=True)
             parser.add_argument("--device", type=str, choices=["cpu", "cuda"], default=None)
             cfg = parser.parse_args(resolve_config_args(sys.argv[2:]))
+            check_compatibility(cfg)
 
             run_config_yaml = f"subcommand: fit\n{parser.dump(cfg)}"
 
@@ -84,6 +86,7 @@ def _run(subcommand: Subcommand):
             parser.add_argument("--config", action=ActionConfigFile)
             parser.add_function_arguments(format, as_positional=False)
             cfg = parser.parse_args(resolve_config_args(sys.argv[2:]))
+            check_compatibility(cfg)
 
             if hasattr(cfg, 'config'):
                 del cfg.config  # pyright: ignore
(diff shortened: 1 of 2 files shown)


morrisoncheney (approved):
lgtm

Plyb (changes requested):
The algorithm here appears to be:
1. recursively get the type arguments from the config
2. match up the arguments with RawT, etc.
3. If there are any two type arguments for the same parameter that aren't the same, fail

Unfortunately, I don't think this will work. This compares type arguments that don't necessarily have to do with each other. For instance, we might have an intervention (`TrainableModel`) that has a different `RawT` from the inner `TrainableModel` it intervenes on (maybe does some kind of preprocessing or something).

This also doesn't take subtyping into account at all, so if a dataset has *more* data than what the model expects (which functionally is fine), this would flag.

PR #402, opened 2026-06-30
40 hyperparam grid search

### LINKED ISSUE #40: Hyperparameter grid search
**Acceptance Criteria**:
- There is a way for a user to specify multiple different values for one or more configuration options
- When multiple configuration options are specified, a separate training run is kicked off for each combination of specified options (this is called a grid search)

JansenNye:
Tested by submitting training runs with the following config files:

config_grid_class.yaml:
```yaml
data:
  class_path: WikitextDataset
  init_args:
    split: train
    head: 50

model:
  class_path: mirror.models.mirror_llama_model.MirrorLlamaModel
  init_args:
    initialization:
      vocab_size: 128256
      hidden_size: 128
      intermediate_size: 340
      num_hidden_layers: 6
      num_attention_heads: 4

slurm:
  job_type: "compute"
  time: "00:10:00"
  gpus_per_node: p100:1

epochs: 1
batch_size: 1
device: cuda

grid:
  model.init_args.initialization:
    - {vocab_size: 128256, hidden_size: 128, intermediate_size: 340, num_hidden_layers: 6, num_attention_heads: 4}
    - {vocab_size: 128256, hidden_size: 256, intermediate_size: 680, num_hidden_layers: 8, num_attention_heads: 4}
  trainer:
    - class_path: TrainerConstructor
      init_args:
        callbacks:
          - class_path: WandbCallback
            init_args:
              log_every_n_steps: 5
    - null
```

config_grid_scalar.yaml:
```yaml
data:
  class_path: WikitextDataset
  init_args:
    split: train
    head: 50

model:
  class_path: mirror.models.mirror_llama_model.MirrorLlamaModel
  init_args:
    initialization:
      vocab_size: 128256
      hidden_size: 128
      intermediate_size: 340
      num_hidden_layers: 6
      num_attention_heads: 4

slurm:
  job_type: "compute"
  time: "00:10:00"
  gpus_per_node: p100:1

epochs: 1
batch_size: 1
device: cuda

grid:
  epochs: [1, 2]
  model.init_args.lr: [1e-3, 1e-4]
```

And confirming that in each case, 4 training runs are correctly submitted.
(Probably `scancel` those runs once you submit them; I actually got priority'd on the last 6 jobs)
Closes #40

### DIFF
--- a/src/mirror/slurm_launcher.py
+++ b/src/mirror/slurm_launcher.py
@@ -6,21 +6,33 @@ from pathlib import Path
 
 from jinja2 import Environment, FileSystemLoader, StrictUndefined
 
-from mirror.slurm_util import SlurmConfig
+from mirror.slurm_util import SlurmConfig, expand_grid, grid_override_args, parse_grid, parse_slurm_config
 from mirror.util import is_login_node
 
 
-def submit_slurm_job(slurm: SlurmConfig, python_args: list[str]) -> None:
-    """Submit a slurm job and exit, if `slurm.job_type` is 'compute' on a login node.
+def submit_slurm_jobs(python_args: list[str]) -> None:
+    """Submit one compute job per `grid` combination (one job if no grid), then exit.
 
     `python_args` is everything after the python script (e.g. `sys.argv[1:]`),
-    starting with the subcommand. Returns normally if no job was submitted;
-    calls `sys.exit` after a successful submission so the caller doesn't have
-    to bail out.
+    starting with the subcommand. No-op when not launching a compute job from a
+    login node, so the caller proceeds to run in-process.
     """
-    if not is_login_node() or slurm.job_type != "compute":
+    if not is_login_node():
         return
 
+    submitted = False
+    for combo in expand_grid(parse_grid(python_args)):
+        args = python_args + grid_override_args(combo)
+        slurm = parse_slurm_config(args)
+        if slurm.job_type == "compute":
+            _submit_one(slurm, args)
+            submitted = True
+
+    if submitted:
+        sys.exit(0)
+
+
+def _submit_one(slurm: SlurmConfig, python_args: list[str]) -> None:
     env = Environment(
         loader=FileSystemLoader(Path(__file__).parent / "templates"),
         undefined=StrictUndefined,
@@ -41,4 +53,3 @@ def submit_slurm_job(slurm: SlurmConfig, python_args: list[str]) -> None:
         )
     job_id = res.stdout.strip().split()[-1]
     print(f"Submitted batch job {job_id}")
-    sys.exit(0)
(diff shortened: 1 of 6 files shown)


morrisoncheney (commented):
Looks mostly good. I think I'm gonna pass this on to Koby and then I'll probably have to be the one to implement the changes we need.

It looks like this won't work if the training run is submitted to be local.
Also, I wonder if we can facilitate the organization and such of the different grids. Minimally, I think printing the job id inside of the slum log would be useful, and with that it would help to print the grid config info somewhere, maybe in the slurm log, or maybe just in the terminal when the job is being submitted.

I think the biggest change needed is that this doesn't really work on local runs and it will just silently ignore the grid (if I am not mistaken).

src/mirror/slurm_launcher.py
```diff
@@ -6,21 +6,33 @@
...
+    submitted = False
+    for combo in expand_grid(parse_grid(python_args)):
+        args = python_args + grid_override_args(combo)
+        slurm = parse_slurm_config(args)
+        if slurm.job_type == "compute":
+            _submit_one(slurm, args)
+            submitted = True
+
+    if submitted:
+        sys.exit(0)
```
Plyb, line 32: If I'm reading this right, we only bail if none of the grid configurations launch a compute job. But if grid search only works on compute jobs anyway, that check is redundant.

Plyb (commented):
Does this mean that the values outside of `grid` in the config are simply ignored/overridden  by `grid`? Can they be omitted if that's the case? Ideally we could omit them if they're not actually used.

Also, agreed on Mo's comment about the local training runs. At the very least, we probably should have a message saying that grid is ignored on non-slurm runs. The alternative I see is to do each grid run *sequentially* on local.

Let's double check this works without a grid at all as well. We don't want to break the existing way of doing things.

JansenNye:
> Does this mean that the values outside of `grid` in the config are simply ignored/overridden by `grid`? Can they be omitted if that's the case? Ideally we could omit them if they're not actually used.
> 
> Also, agreed on Mo's comment about the local training runs. At the very least, we probably should have a message saying that grid is ignored on non-slurm runs. The alternative I see is to do each grid run _sequentially_ on local.
> 
> Let's double check this works without a grid at all as well. We don't want to break the existing way of doing things.

Yes, they can be omitted.

I think I'll enforce an error if you try to run a grid submission locally.
 
And it does still work without a grid. `expand_grid({})` returns `[{}]`, so a config without a grid section goes through the identical code path as before

src/mirror/slurm_launcher.py
JansenNye: The job-type check itself has to stay because it's what lets a plain no-grid local run skip submission, but repeating that check for every combination is pointless, so I restructured to just check beforehand

PR #409, opened 2026-07-01
Implement config file language validation

### LINKED ISSUE #61: Language server validation for config files
See if we can get some nice vscode language support for the config files. Possibly the first or second option in this conversation could work: https://gemini.google.com/share/c8d325207d82

**Acceptance Criteria**:
- Either:
  - When opening a configuration file in vscode, I get type validation based on what the training script takes in
    - There is minimal code duplication (i.e. we shouldn't specify the schema in more than one place if possible)
  - There is a note written up about why we can't/shouldn't do this

JansenNye:
Tested by confirming that valid config files raise no errors, and editing one to have `epochs: one` instead of `epochs: 1` and confirming that it raises an error. 
Notes:
- Make sure you guys install the YAML extension by Red Hat to allow this to work
- You'll need to paste the following at the top of a config file for it to get validated:
```# yaml-language-server: $schema=../../schemas/<subcommand>.schema.json```
With `<subcommand>` being `fit`/`eval`/`format`/`infer` - whatever subcommand the config file is for

Closes #61

### DIFF
--- a/schemas/eval.schema.json
+++ b/schemas/eval.schema.json
@@ -1,0 +1,79 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "type": "object",
+  "properties": {
+    "config": {},
+    "metrics": {},
+    "checkpoint_path": {
+      "type": [
+        "string",
+        "null"
+      ]
+    },
+    "slurm": {
+      "type": "object",
+      "properties": {
+        "job_type": {
+          "enum": [
+            "compute",
+            "local",
+            "local-download"
+          ]
+        },
+        "time": {
+          "type": "string"
+        },
+        "nodes": {
+          "type": [
+            "integer",
+            "null"
+          ]
+        },
+        "ntasks_per_node": {
+          "type": [
+            "integer",
+            "null"
+          ]
+        },
+        "gpus_per_node": {
+          "type": [
+            "string",
+            "null"
+          ]
+        },
+        "mem_per_cpu": {
+          "type": "string"
+        },
+        "output": {
+          "type": "string"
+        },
+        "open_mode": {
+          "type": "string"
+        },
+        "signal": {
+          "type": "string"
+        },
+        "requeue": {
+          "type": "boolean"
+        },
+        "qos": {
+          "type": [
+            "string",
+            "null"
+          ]
+        }
+      },
+      "additionalProperties": false
+    },
+    "model": {},
+    "strategy": {},
+    "device": {
+      "enum": [
+        "cpu",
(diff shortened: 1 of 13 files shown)

### CODE REVIEW

**Review by morrisoncheney — APPROVED** [review]
Looks good to me!
One note: It worked in correcting some but not all mistakes on configs. e.g. it did not correct mistakes in the strategy for inference. Looks like anything we want it to be able to ensure is accurate needs to be added to the .json files of this PR. I guess maybe we could just add tickets for such things as they come up?

File: schemas/eval.schema.json [hunk 1]
```diff
@@ -0,0 +1,79 @@
...
+        },
+        "qos": {
+          "type": [
+            "string",
+            "null"
+          ]
+        }
+      },
+      "additionalProperties": false
+    },
```
Plyb, line 67: This (and some of the other properties) are duplicated between schemas. Is there a way to share them?

EDIT: oh! I see! these are programmatically generated. Cool!

src/mirror/subcommands.py
```diff
@@ -99,5 +99,6 @@ def infer(
 def format(
         data: MirrorDataset,
         formatter: MirrorFormatter,
+        slurm: SlurmConfig = SlurmConfig(),
```
Plyb, line 102: Why was this added?

src/mirror/config_schema.py
```diff
@@ -0,0 +1,95 @@
...
+        "properties": _build_props(tree),
+        "additionalProperties": False,
+    }
+
+
+def write_schemas(out_dir: Path = _SCHEMAS_DIR) -> None:
+    out_dir.mkdir(exist_ok=True)
+    for subcommand in SUBCOMMANDS:
+        schema = generate_schema(build_parser(subcommand))
+        (out_dir / f"{subcommand}.schema.json").write_text(json.dumps(schema, indent=2) + "\n")
```
Plyb, line 91: The types here are pretty loose (lots of plain `dict`s for instance). Can we tighten them up at all? Optional, since reflection code like this tends to be pretty hard to type well.

Plyb (changes requested):
Super cool! Just a couple thoughts. Plus: I wonder if it might be good to have a commit hook/GH action to make sure the schema files stay in sync. Probably for a follow up ticket though

src/mirror/subcommands.py
JansenNye: Format configs actually do have a `slurm: ` section that's read at submission time by `parse_slurm_config` apparently

Plyb (approved):

